PR #505, opened 2026-06-10
fix(images): strip EXIF/GPS metadata from uploaded images

jagduvi1:
## Problem

Audit finding (High, GDPR — verified in both the 2026-06-05 and 2026-06-10 audits): uploaded image originals are stored **byte-for-byte with EXIF intact** and served unauthenticated under `/api/uploads` with a 1-year cache. Phone photos embed GPS coordinates — for bottle photos that is **where the owner stores their wine**. The location leaked to:

- anyone holding the URL (no auth on the static mount),
- shared-cellar members during the pre-processing window (`processedUrl || originalUrl` fallback),
- the public wine image when an unprocessed image is approved,
- the user's own GDPR export.

The background-removed PNG was already clean (rembg re-encodes via PIL); the raw original was the leak.

## Fix

- **New `services/imageSanitizer.js`** — re-encodes uploads in place with `sharp`, which drops EXIF/XMP/IPTC by default. `.rotate()` bakes the EXIF orientation into the pixels first so photos don't render sideways; the ICC profile is kept (`keepIccProfile()`) to avoid colour shift on wide-gamut phone photos.
- **Applied to both user-photo upload paths**: bottle images (`POST /api/images/upload`) and wine-list logos (`POST /api/wine-lists/:id/logo`). Runs after authorization checks, before the file is referenced anywhere.
- **Defense-in-depth bonus**: `limitInputPixels: 8000×8000` makes the previously *fail-open* dimension check fail **closed** at decode time for formats the header parser can't read (extended VP8X WebP, JPEGs with >64KB header segments) — closes the rembg pixel-bomb edge on this path. Corrupt files that pass the magic-byte check now get a clean 400 instead of flowing onward.
- Sibling-consistency fix while touching the lines: the invalid-`wineDefinitionId` branch now unlinks the uploaded file like every other validation branch (audit Low).

## Migration for existing files

```bash
docker exec cellarion-backend node src/scripts/strip-exif-images.js
```

Idempotent, no DB access, filenames/URLs unchanged. Sweeps `originals/` and `wine-list-logos/`; `processed/` is skipped (already clean).

## Testing

- ✅ New unit tests (`imageSanitizer.test.js`, 5 cases): EXIF+GPS stripped for JPEG/PNG/WebP, orientation baked in (dims swap verified), corrupt file rejected, unknown extension rejected untouched.
- ✅ Full backend suite: **673/673** across 34 suites.
- ✅ Docker smoke test: built the image (sharp installs clean on node:20-alpine), uploaded a GPS-tagged JPEG via the real API as the demo user — served original came back **EXIF-free**, correct format/dimensions.
- ✅ Migration script run in the container against 106 pre-existing files: `106 stripped, 0 failed`.

## Notes

- `sharp` added as a backend prod dependency (~32MB installed) — prebuilt binaries, no build step, the clear right tool per the dependency philosophy (correct EXIF/orientation/ICC handling across 3 formats is not a few lines of native code).
- JPEG/WebP re-encode at quality 90; PNG lossless.
- Does **not** change the unauthenticated `/api/uploads` static mount (separate audit finding — the served bytes are now clean, the access-control question stays open).


jagduvi1:

PR #506, opened 2026-06-10
refactor(perf): scale phase 1 — indexes, payload projections, bundle splits

jagduvi1:
## What

First batch from [OPTIMIZATION_AUDIT_2026-06-10.md] — the cheap, additive scalability wins. No behavior changes except where noted.

### Database — 10 new indexes
| Model | Index | Why |
|---|---|---|
| Bottle | `{cellar, createdAt: -1}`, `{user, createdAt: -1}` | The default newest-first list sort had no supporting index → blocking in-memory SORT over every bottle of the cellar/user on **every page view** (the status filter is a `$nin` range, so equality+sort-order indexes are the correct shape) |
| User | `{refreshTokenHash}` sparse | Every `/api/auth/refresh` (one per session ≈15min) was a **full collection scan** |
| Discussion | newest tab, most-replies/trending tab, `{wineDefinition, isPinned, lastActivityAt}` | Public forum tabs + the crawler-visited wine-page discussions panel blocking-sorted the whole collection |
| WineDefinition | `{country\|region\|grapes\|type, name}` | Public taxonomy discovery pages sorted entire country/type wine sets in memory; the `grapes` prefix also serves the taxonomy count queries |

Verified live: all 10 built by mongoose autoIndex on boot.

### Payload — `WINE_POPULATE_LIST`
List/stats paths (bottles list, cellar views, history, stats) now populate `wineDefinition` **excluding** `aiProfile` (several KB of prose per wine), `normalizedKey`, `lwin`, `productNumber(s)`, `createdBy`. Verified by grep that no list consumer reads them; detail routes (`GET /api/bottles/:id` etc.) keep the full populate — the bottle page renders `aiProfile`. Smoke-verified: 15 vs ~21 keys per populated wine, card fields intact.

### Slot picker — functional fix 🐛
The rack/room slot pickers sent **every placed bottle ID in the query string** (`?exclude=id,id,…`) — past ~300-600 placed bottles that overflows the URL/header limit and "place a bottle" breaks entirely. New `?excludePlaced=1` resolves placed IDs server-side from the cellar's racks. The old `exclude` param still works for other callers (AddBottle flow).

### Bundle (measured before/after with `vite build`)
- **Entry: 161.9 → 133.9 kB gzip** — Swedish locale now lazy-loads (en stays bundled as the fallback; sv users may see a one-frame English flash on first cold load)
- **three.js (249.5 kB gzip)** no longer loads on the everyday 2D racks page — lazy behind the 3D toggle
- **tiptap composer (119.6 kB gzip)** no longer ships to read-only/anonymous discussion visitors — lazy at the form render sites
- `BottleCard` memoized — CellarDetail's accumulated load-more list re-rendered every card per search keystroke
- Verified in `dist/`: both chunks moved from static import preambles to dynamic-import dep maps

### Notifications
- New lightweight `GET /api/notifications/unread-count` (one indexed count), polled at **60s** instead of the full list at 30s, **skipped while the tab is hidden**; the full list refetches only when the count changes. Per-tab budget impact drops from 30 req/15min to ≤15 cheap req/15min.
- Side-effect fix: the badge previously undercounted (computed from the 30 most recent rows only — audit finding from 2026-06-05); it now counts all unread rows.

### Infra (sized for the current 4GB Hetzner host)
- Mongo WiredTiger cache **0.5 → 0.25 GB**: the cache equalled the whole 512M container limit — an OOM-kill waiting to happen under load
- Backend **384 → 640 MB**: stopgap headroom for full-cellar hydration until those endpoints move to aggregation (phase 2)

## Testing
- ✅ Backend 668/668 (33 suites), frontend 300/300 (9 files)
- ✅ `vite build` clean; chunk movement verified in dist output
- ✅ Docker smoke test against the rebuilt stack: login → `unread-count` works → bottles list has heavy fields stripped + card fields present → `?excludePlaced=1` returns only unplaced bottles (5 of 9 with demo data) → all 10 indexes present in mongo
- ✅ Mongo recreated and running with the 0.25 WT cache flag

## Not in this PR (phase 2+)
Cellar `group=1` full-hydration → aggregation; stats → pipelines; taxonomy N+1 counts; reply fan-out batching; chunked Meili full-sync; job hygiene; worker split/replicas.

### DIFF
--- a/backend/src/config/constants.js
+++ b/backend/src/config/constants.js
@@ -16,6 +16,20 @@ const WINE_POPULATE = [
   { path: 'pendingWineRequest', select: 'wineName producer' }
 ];
 
+// List/stats variant of WINE_POPULATE: excludes the heavy fields no list view
+// or aggregate reads (aiProfile alone is several KB of prose per wine). A page
+// of 30 bottles ships the same card data at a fraction of the payload. Detail
+// routes (GET /api/bottles/:id etc.) keep the full WINE_POPULATE — the bottle
+// page renders aiProfile.
+const WINE_POPULATE_LIST = [
+  {
+    path: 'wineDefinition',
+    select: '-aiProfile -normalizedKey -lwin -productNumber -productNumberShort -createdBy',
+    populate: ['country', 'region', 'grapes']
+  },
+  { path: 'pendingWineRequest', select: 'wineName producer' }
+];
+
 // ─── Import thresholds ───────────────────────────────────────────────────────
 
 // Composite similarity score at or above which a match is considered exact
@@ -70,6 +84,7 @@ module.exports = {
   CONSUMED_STATUSES,
   MS_PER_DAY,
   WINE_POPULATE,
+  WINE_POPULATE_LIST,
   IMPORT_EXACT_THRESHOLD,
   IMPORT_FUZZY_THRESHOLD,
   MAX_IMPORT_SIZE,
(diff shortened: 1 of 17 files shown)


jagduvi1:

PR #507, opened 2026-06-10
refactor(perf): scale phase 2a — DB-side grouping, stats cache, taxonomy aggregation

jagduvi1:
## ⚠ Stacked on #506

Base branch is `refactor/scale-phase1` — **merge #506 first**, then retarget/rebase this onto main (I can do the rebase on request). Stacked because both touch the same code in `cellars.js`.

## What

The structural request-path work from the optimization audit — kills the three biggest per-request scalability problems.

### 1. Cellar page grouping → MongoDB aggregation (the #1 finding)
`GET /api/cellars/:id?group=1` — the **default cellar page** — previously set `canPaginateInDb=false`, loaded **every bottle in the cellar** with full populate (≈6 docs per bottle), grouped them in JS, and sliced 30 groups. That ran on every page view *and every search keystroke*: a 2,000-bottle cellar ≈ 12k documents and 25-50MB heap per request.

New `loadGroupedBottlePage()`:
- groups by (wine, vintage, bottleSize) **inside MongoDB** (`$sort → $group → $sort → $skip/$limit`, `allowDiskUse`),
- then populates **only the returned page's members** (≤ ~30 groups),
- group keys + member order are **byte-identical** to the JS algorithm — `$first`/`$push` follow the pipeline's `$sort`, and empty-string vintage/size normalize exactly like the JS path (`'' → 'NV'/'750ml'`).
- Applies to the hot path only (no filters, DB-sortable sort). Filtered/maturity/minRating views keep the existing paths — they need cross-document context the DB can't compute (rating-scale normalization, maturity profiles).
- Aggregation pipelines bypass Mongoose casting, so `cellar`/exclude ids are cast to ObjectId explicitly (the audit found a real bug of this exact class in priceWarnings).

### 2. Stats overview — server-side cache with count-probe invalidation
`GET /api/stats/overview` loads the user's entire bottle set (active + consumed, populated) per request. Now cached **2 minutes per user**, with an invalidation probe of two cheap indexed `countDocuments` — add/delete/consume/undo all change one of the counts and bust the cache immediately; only pure field edits (price/rating tweaks) can be served ≤2min stale. Bounded cache (5k entries, cleared on overflow).

### 3. Taxonomy counts → one `$group` per endpoint
`/api/taxonomy/{countries,regions,grapes}` ran **one `countDocuments` per taxon, sequentially** — the grapes list was hundreds of full-registry scans per request (no `grapes` index until #506). Each list is now a single aggregation (`$unwind + $group` for grapes) + a 10-minute server-side cache (clients already get `max-age=3600`).

## Testing
- ✅ Backend 668/668
- ✅ Docker smoke with **seeded parity fixture**: 3× same wine/vintage/size (multi-member group) + different vintage + 1.5L magnum + NV bottle → aggregation output vs the JS grouping algorithm replicated client-side: **6/6 groups, keys match, member order matches**
- ✅ Stats: consecutive calls byte-identical (cached, 8-11ms); taxonomy lists consistent with counts present

## Deferred (noted in audit)
- `/api/bottles` search via Meilisearch: the bottles index has no user field — needs a full re-index to add; separate change.
- `/history` fallback and Meili-path grouped views (rarer paths) keep in-memory behavior.

### DIFF
--- a/backend/src/routes/admin/taxonomy.js
+++ b/backend/src/routes/admin/taxonomy.js
@@ -17,6 +17,19 @@ const router = express.Router();
 // All routes require admin role
 router.use(requireAuth, requireRole('admin'));
 
+// Public taxonomy lists are served from a 10-minute server-side cache —
+// drop it after any successful admin mutation so renames/deletions show up
+// immediately instead of after the TTL.
+const { clearTaxonomyListCache } = require('../taxonomy');
+router.use((req, res, next) => {
+  if (req.method !== 'GET') {
+    res.on('finish', () => {
+      if (res.statusCode < 400) clearTaxonomyListCache();
+    });
+  }
+  next();
+});
+
 // ===== COUNTRIES =====
 
 // GET /api/admin/taxonomy/countries - List all countries
--- a/backend/src/routes/cellars.js
+++ b/backend/src/routes/cellars.js
@@ -30,6 +30,86 @@ function getUserColor(cellar, userId) {
   return entry?.color || null;
 }
 
+// Group key parts default like the JS grouping path: missing/empty vintage →
+// 'NV', missing/empty bottleSize → '750ml'. $ifNull alone misses empty strings.
+function groupPartExpr(field, fallback) {
+  return {
+    $let: {
+      vars: { v: { $ifNull: [`$${field}`, fallback] } },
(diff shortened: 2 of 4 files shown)


jagduvi1:

PR #508, opened 2026-06-10
refactor(perf): scale phase 2b — batched fan-out, chunked Meili sync, job hygiene

jagduvi1:
## What

The jobs-and-sync half of Phase 2 from the optimization audit. Branches off main directly — **independent of #506/#507** (disjoint files), mergeable in any order.

### 1. Reply notification fan-out → one batch
One forum reply previously fired **~3 Mongo queries + per-subscription web-push HTTPS calls per watcher**, all un-awaited inside the request handler — a 500-watcher thread ≈ 1,500 concurrent queries stampeding the connection pool.

New `createNotifications(items)` in the notifications service:
- one `insertMany` for all in-app rows,
- **one** prefs query + **one** push-subscription query for all recipients (`$in`),
- push delivery through the existing `runConcurrent` util, capped at 10 in-flight,
- same per-category preference gating (extracted to `pushAllowedFor`, identical semantics including the legacy permissive fallback),
- `createNotification` now delegates to a batch of one → every other caller keeps byte-identical behavior.

### 2. Meilisearch full syncs → cursor + chunks, non-blocking, fail-closed
- `fullSync`/`fullSyncBottles`/`fullSyncDiscussions` previously loaded **entire collections into one array and one HTTP payload** — OOMs Node and exceeds Meili's ~100MB payload cap around 200k bottles. Now streamed via Mongoose cursor into 2,000-doc `addDocuments` chunks (discussions: per-chunk `$in` reply fetch — batched, never the whole replies collection in memory).
- Initial syncs no longer block `app.listen` — a multi-minute catalog upload used to gate boot **and the container healthcheck**, making first-boot/recovery deploys fail at scale. Search may briefly return partial results during a first-boot sync (only then).
- A failed index-stats check now **skips** the sync (fail closed) instead of re-uploading the whole catalog exactly when Meilisearch is struggling. `MEILI_FORCE_REINDEX=1` still forces.

### 3. Job hygiene
- **communityPriceJob**: `allowDiskUse(true)` + cursor iteration — the nested `$push` stages blow Mongo's 100MB stage limit on a big price universe, and `rows` held every price in the system in Node memory.
- **drinkWindowNotifier first-run seed**: collected into one `bulkWrite` per user instead of one awaited `updateOne` per bottle (the seed touches every bottle in the system — hours of sequential writes at scale).
- **cleanupOrphanedImages**: async `fs.promises` (the sync walk froze the event loop for the whole directory scan, hourly), and 🐛 **fixes the audit-verified crash**: `originalUrl` is null for approved/rejected images, `path.basename(null)` threw, and the swallowed error silently disabled the orphan sweep forever once any image had been moderated.

## Testing
- ✅ Backend 668/668
- ✅ Docker smoke: admin creates discussion → user replies → **batch notification delivered end-to-end** (correct title/message, unreadCount 0→1)
- ✅ Boot logs confirm: populated indexes skip sync; the discussions sync ran through the new chunked path; and `[cleanup] Removed 1 orphaned original files` — **the sweep completed and reaped an orphan for the first time** (it had been crashing every hour before)

## Deferred (Phase 3 / audit backlog)
Embedding-job Voyage batching + build-then-swap Qdrant rebuild; per-user drink-window loop parallelism; outbox/queue for email+push; scheduler worker split.

### DIFF
--- a/backend/src/routes/discussions.js
+++ b/backend/src/routes/discussions.js
@@ -817,40 +817,47 @@ router.post('/:idOrSlug/replies', requireAuth, async (req, res) => {
       maybeEmail(quotedAuthorId, 'communityReply', 'quote');
     }
 
-    const mentionedIds = await extractMentions(cleanBody, req.user.id);
-    for (const uid of mentionedIds) {
-      if (notified.has(uid)) continue;
-      notified.add(uid);
-      notificationItems.push({
-        userId: uid,
-        type: 'discussion_mention',
-        title: `${replierName} mentioned you`,
-        message: `In "${discussion.title}"`,
-        link,
-        category: 'communityMention',
-      });
-      maybeEmail(uid, 'communityMention', 'mention');
-    }
+    // Mention/watcher resolution is best-effort: the reply is already saved,
+    // so a failure here must not 500 the request or lose the OP/quote
+    // notifications already collected above.
+    try {
+      const mentionedIds = await extractMentions(cleanBody, req.user.id);
+      for (const uid of mentionedIds) {
+        if (notified.has(uid)) continue;
+        notified.add(uid);
+        notificationItems.push({
+          userId: uid,
+          type: 'discussion_mention',
+          title: `${replierName} mentioned you`,
+          message: `In "${discussion.title}"`,
+          link,
+          category: 'communityMention',
+        });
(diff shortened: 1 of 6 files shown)


jagduvi1:

PR #511, opened 2026-06-11
fix(wine-lists): path traversal, cross-tenant bottle exposure, logo file orphans

jagduvi1:
Security and hygiene fixes for the Wine List feature (two findings also flagged in the June 10 audit).

## Fixes

- **Cross-tenant bottle exposure (audit: Medium):** `loadBottleMap` queried `Bottle.find({ _id: { $in: ids } })` unscoped while entry bottle IDs are client-supplied — another tenant''s bottle name/producer/price could be rendered into a (public) PDF. The loader is now shared in `services/wineListData.js` and scoped to the list''s cellar, used by both the authed and public routes.
- **Path traversal via `branding.logoUrl` (audit: Medium):** `PUT /api/wine-lists/:id` mass-assigned `branding`, letting a client-supplied `logoUrl` flow into `path.join(''/app/uploads'', logoUrl)` → `doc.image()`. The stored, server-generated value is now always preserved on PUT, and the PDF renderer additionally validates the `wine-list-logos/<uuid>.<ext>` pattern before touching the filesystem.
- **Logo file orphans:** replaced logos are deleted after a successful re-upload, list deletion removes the logo file, and the GDPR account purge unlinks all of a user''s wine-list logos before deleting the docs.
- Removed the dead "verify owner''s plan" comment in the public route (all tiers have full feature access) and corrected the `requireCellarOwner` doc comment (wine lists are owner-only, matching the UI gating).

## Testing

- `cd backend && npm test` — 675 tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #512, opened 2026-06-11
feat(wine-lists): re-key entries to wine+vintage+size, per-wine glass pricing

jagduvi1:
Stacked on #511. The core redesign that makes the Wine List behave like a restaurant wine list.

## The problem

Entries referenced **physical bottles**. A cellar with 12 bottles of the same Barolo showed 12 identical rows in the editor, quick-start put all 12 on the menu, and consuming the one specific referenced bottle silently removed the wine from the published PDF even with 11 still in stock.

## The redesign

Entries now identify a **wine on the menu**: `WineDefinition + vintage + bottle size` (magnums list as their own line, as on a real list).

- Duplicates collapse into one line; the line survives bottle consumption while stock remains.
- Out-of-stock is now an explicit `layout.hideOutOfStock` toggle instead of silent per-bottle disappearance.
- The dashboard keeps zero-stock wines visible as restock warnings (previously they were silently skipped).

## By the glass, per wine

- `entry.byGlass` marks a wine as served by the glass; ticking it instantly pre-fills a suggested price from the list-level rule (`price / glassesPerBottle × (1 + markup%)`, rounded to nearest 1/5/10).
- Prices the user types are flagged `glassPriceManual` and are never overwritten by recalculation; suggested prices follow bottle-price changes live.
- Optional **"Wines by the Glass" lead section** (localized in all 6 PDF languages) via `layout.glassSectionFirst`.
- The all-or-nothing `layout.showGlassPrice` flag and its overwrite-everything "Calculate" button are retired.

## Supporting changes

- `GET /api/wine-lists/cellar-wines` — cellar bottles aggregated per wine+vintage+size with stock and avg purchase price (the editor''s picker; also removes the old 500-bottle cap).
- `GET /api/wine-lists/:id` returns `resolvedWines` so entries whose wine is out of stock stay editable.
- PDF prints non-standard formats explicitly ("(1.5L)") and adds up to three grape names to the detail line.

## ⚠️ Migration required

`docker exec cellarion-backend node src/scripts/migrate-wine-list-entries.js` after deploy. Idempotent; maps each bottle entry to its wine+vintage+size, dedupes, and preserves exactly the glass prices the published PDF showed before. Old bottle-keyed entries do not render until migrated.

## Testing

- 15 new unit tests (section building, entry resolution, migration mapping); backend 690 + frontend 300 pass.


jagduvi1:
**Code audit follow-up**: commit 9e10350 fixes the four audit findings in this PR''s scope:

1. **Admin wine merge orphaned wine list entries** — entries reference WineDefinition directly now, so both merge paths re-point `entry.wine` refs (previously the merged-away wine silently vanished from published menus; bottle-keyed entries used to follow `Bottle.updateMany` automatically).
2. **Migration is now reversible and stock-aware**: every list is backed up to `winelists_premigration_backup` before rewriting; consumed-bottle entries are kept when the wine still has active stock and dropped when sold out; `layout.hideOutOfStock` is enabled on migrated lists so sold-out wines stay off published menus exactly as before.
3. **Hand-entered glass prices are protected**: `glassPriceManual` is set whenever a glass price exists (the old system had no suggestion engine, so every stored price was manual) — previously prices hidden behind the old display toggle could be silently overwritten by the new pricing rule.
4. **Glass-only prices render**: a byGlass wine with no bottle price no longer shows a blank price column on the PDF. Also: dropped pending-wine-request entries are logged loudly with their prices, and `resolveEntry` now uses the canonical `entryKey()`.

PR #513, opened 2026-06-11
feat(wine-lists): public web menu, live preview, autosave, reordering, i18n

jagduvi1:
Stacked on #512 (which stacks on #511). The guest-facing and editor-UX layer.

## Guest-facing web menu

- New public page **`/menu/:token`** renders the published list as a mobile-first web menu (same four color schemes as the PDF), with the PDF one tap away. Backed by `GET /api/wine-lists/public/:token` returning the list as JSON with inventory data stripped.
- **QR codes on printed PDFs now point at the web menu** instead of forcing a PDF download onto guests'' phones.
- Public rate limit raised 30 → 120 req/15min — a restaurant''s guests share one NAT''d IP.

## Editor

- **Live Preview tab** that reflects unsaved state, via a client-side section builder (`utils/wineListSections.js`) and a `WineListMenu` component shared with the public page.
- **Autosave**: 1.5s debounce, serialized (one request in flight), no retry-hammering after a failure until the user edits or presses Save; Saved/Unsaved/Saving indicator in the header plus a `beforeunload` guard. Manual Save remains as an instant flush.
- **Reordering**: up/down controls for custom sections and for wines within a section (`sortOrder` finally has a UI).
- Share tab shows both the public menu link and the PDF link with copy buttons.

## Management

- `POST /api/wine-lists/:id/duplicate` + a Duplicate button on list cards ("Spring Menu" → seasonal copies; share token and published state are not copied).
- New lists inherit the user''s preferred currency (symbol map added to `config/currencies.js`) instead of defaulting to `$`.
- `WineLists`, `WineListEditor` and the CellarDetail links are wired to the existing `wineLists.*` translation keys — a full en+sv key set existed in the locale files but was never used — plus new en/sv keys for the glass-pricing and menu features. Guest-facing strings on `/menu` follow the **list''s** language (all 6).

## Testing

- 6 new Vitest tests for the section builder; frontend 306 + backend 690 pass; `vite build` clean.


jagduvi1:
**Code audit follow-up**: commit 4438fd8 fixes the audit findings in this PR''s scope:

- **Duplicate shared the logo file**: duplicating a list now copies the logo file under a fresh name — deleting or re-uploading on either list no longer breaks the other.
- **Autosave flushes on unmount**: in-app router navigation doesn''t fire `beforeunload`, so edits made within the 1.5s debounce window were silently lost; a pending payload is now saved fire-and-forget when the editor unmounts. `payloadJson` is memoized (was re-stringified on every render).
- **Editor/menu order stays in sync**: custom-mode `sortOrder` is renumbered on every add/remove/move (select-all and toggle could produce colliding sortOrders, making the rendered order diverge from the editor).
- **QR fallback**: when `FRONTEND_URL` is unset the QR points at the self-referencing PDF URL instead of a `/menu` route the API host may not serve.
- Glass-only prices render on the web menu; public JSON endpoint uses `.lean()`.

PR #595, opened 2026-07-07
feat: silent image delete with duplicate handover in by-wine curation

jagduvi1:
## Problem
The by-wine **Remove** action reused the moderation *reject*: it notified the uploader (wrong for housekeeping), left a rejected tombstone, and had two real data hazards:
1. **Starred images went dangling** — if a user had the removed image starred as their bottle's preferred image, the reference dangled and their choice was silently lost, even when a byte-identical duplicate survived.
2. **Shared files could be deleted** — import dedup can point two image records at the *same file*; the raw unlink deleted it for the surviving record too.

## Fix
New **`DELETE /api/admin/images/:id`** (admin-gated, audit-logged) used by the by-wine Delete button:
- **Silent** — no uploader notification (per product decision: duplicate cleanup is not a judgement on the upload).
- **Starred handover** — bottles whose `defaultImage` was the removed record are repointed to a **byte-identical surviving copy** (same `contentHash`, same uploader preferred), or cleared to fall back to the wine image. Never dangling.
- **Official handover** — if the removed image was the wine's official image, official status transfers to the identical survivor (mirrors set-official) instead of leaving the wine imageless.
- **Reference-safe file removal** via the existing `unlinkImageFiles`; the record is hard-deleted (no tombstone).

The **status-queue Reject keeps its moderation semantics** (notification + tombstone) but gains the same reference-safe unlink + starred-image cleanup, fixing the identical hazards there.

## Testing
E2E against the running backend (throwaway data, cleaned up): 14/14 assertions across 3 scenarios — starred repoint (2 bottles), official transfer + survivor promotion, deleted file gone + survivor file intact, **no notification created**, shared-file survival, no-survivor case clears to null.

## Docker / compose impact
Code-only — no compose changes.

### DIFF
--- a/backend/src/routes/admin/images.js
+++ b/backend/src/routes/admin/images.js
@@ -5,7 +5,7 @@ const WineDefinition = require('../../models/WineDefinition');
 const Bottle = require('../../models/Bottle');
 const searchService = require('../../services/search');
 const fs = require('fs');
-const { reprocessAllImages, safeUploadPath } = require('../../services/imageProcessor');
+const { reprocessAllImages, safeUploadPath, unlinkImageFiles } = require('../../services/imageProcessor');
 const { logAudit } = require('../../services/audit');
 const { createNotification } = require('../../services/notifications');
 const { incrementCred } = require('../../utils/cellarCred');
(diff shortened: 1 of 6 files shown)


jagduvi1:
Added `backend/src/scripts/cleanup-duplicate-images.js` — one-time bulk cleanup of the existing duplicate backlog (same handover rules as the DELETE endpoint; dry-run by default, `--apply` to execute). Already run on the local dataset: **12 groups, 29 duplicate records removed**, re-run finds 0, zero dangling starred refs. **Run on prod after deploy:** `docker exec cellarion-backend node src/scripts/cleanup-duplicate-images.js --apply` (dry-run first if you want to review).

PR #668, opened 2026-07-10
feat: cellar climate monitoring — ESP32 sensors, open ingest API, charts, alerts

jagduvi1:
Cellar climate monitoring: sensor devices (the ESP32 reference kit — or **any HTTP client**) post temperature + humidity readings into Cellarion; the cellar page shows current values and history charts, and out-of-range / sensor-offline conditions notify through the normal channels.

**Standalone by design — no Home Assistant required.** The ingest API is an open, documented contract (`docs/climate-monitoring.md`); the ESPHome firmware template in `firmware/` is just client #1. An HA `rest_command` recipe is included in the docs for people who already run HA.

## Backend

- New `climate` token scope in the default-deny allowlist — a device token can reach **only** `POST /api/climate/ingest`. Device identity is resolved *from the token* (1:1 `ClimateDevice` ↔ `ApiToken`), so payloads carry no spoofable device id; revoking the token cuts the device off.
- `ClimateReading` is a MongoDB **time-series collection** (verified in Docker: `type: timeseries`, granularity `minutes`, TTL 730 days via `CLIMATE_RETENTION_DAYS`).
- Ingest validation: 1–100 readings/request, value bounds (−30…60 °C / 0…100 %RH), timestamp window (48 h backfill … 5 min future), channel key regex, channels auto-register up to 16/device, per-channel calibration offsets applied at write time.
- **Two-layer rate limiting** (per discussion): per-token **60 requests / 15 min** and **20,000 readings / device / UTC day** (`CLIMATE_MAX_READINGS_PER_DAY`) — the daily quota bounds worst-case storage abuse that request-counting alone can't (batched requests). Quota-blocked devices still count as "seen" so they aren't falsely flagged offline.
- Alerts with hysteresis: breach must persist 15 min before notifying, 6 h cooldown per channel, recovery notifies only when its breach did; sensor-offline detection in the existing 15-min cron, back-online recovery at ingest. New notification types: `climate_alert`, `climate_recovered`, `climate_offline`.
- Cellar-facing endpoints: `current` (last-value cache, no readings query), `readings` (server-side `$dateTrunc` buckets: 5 min/1 h/6 h/1 d for 24 h/7 d/30 d/1 y), owner-only threshold config (defaults 8–16 °C, 45–80 %RH, offline 60 min).
- GDPR: `ClimateDevice` registered in `userDataRegistry.js` (purges devices + readings, exports both, capped + `_truncated`); `ClimateReading` in EXCLUDED with rationale (purged via the device entry). Audit on device create/update/delete — never on readings.

## Frontend

- **Settings → Climate devices**: create (password-confirmed, token shown once — mirrors the API-tokens card), assign to cellar, channel labels + calibration offsets, delete.
- **Cellar page → Overview**: climate card with per-channel current values, alert coloring, online/offline badges, 60 s refresh — renders nothing when the cellar has no devices.
- **History modal** (lazy chunk): recharts, temperature and humidity as **separate charts** (single axis each), threshold guide lines, 24 h/7 d/30 d/1 y ranges, series colors follow the app's existing chart palette.
- i18n: en + sv. ⚠️ **Swedish strings are machine-authored — please review before release.**

## Firmware + docs

- `firmware/cellarion-climate.yaml`: ESPHome template for the decided zero-solder build (2× DS18B20 on one 1-Wire pin + 2× SHT31 on I2C = 4 temp + 2 RH channels), Wi-Fi captive-portal fallback, OTA, LAN status page, 5-min POST interval. `secrets.yaml` is gitignored (holds Wi-Fi creds + token); `.bin` artifacts ignored too.
- `docs/climate-monitoring.md`: the public API contract (stability promise: additive-only), curl/Python/Home-Assistant recipes, alert semantics, retention, self-hosting notes.

## docker-compose impact

**None.** No new service, port, or required env var. The time-series collection is created by Mongoose at boot on the existing Mongo 7. Optional env (safe defaults): `CLIMATE_RETENTION_DAYS` (730), `CLIMATE_MAX_DEVICES_PER_USER` (5), `CLIMATE_SUGGESTED_INTERVAL_S` (300), `CLIMATE_MAX_READINGS_PER_DAY` (20000). Nothing to mirror into the prod compose file.

## Testing

- Backend: 83 suites, **1,326 tests green** (new: alert-hysteresis state machine, ingest validation + daily quota, device CRUD/token-cleanup, allowlist scope pinning; GDPR registry completeness test forces the new model registration).
- Frontend: 28 suites, **640 tests green** (new: api/climate contract tests); `vite build` clean.
- **Docker e2e smoke: 22/22 PASS** — login → create device → default-deny check (device token 403 on `/api/cellars`, JWT 403 on ingest) → ingest with a deliberate out-of-bounds rejection → 48 h backfill → bucketed series → threshold config + validation → channel label/calibration (offset verified applied at ingest) → device delete → revoked-token 401 + readings cascade. Time-series collection type/TTL verified in mongosh.

## Decisions taken (flagged for review)

Shipped defaults for the plan's open questions — all env- or user-tunable: 5-min advisory cadence (min 60 s), 730-day retention, device cap 5/user, thresholds 8–16 °C / 45–80 %RH, **no plan gating** in this MVP.

## Follow-ups (not in this PR)

Phase 0 hardware spike in the real cellar (~€30–40 kit per `firmware/README.md`); Phase 2: rack-view climate overlay, climate data in cellar export; Phase 3 bundle work per `CLIMATE_MONITORING_PLAN_2026-07-10.md`.


jagduvi1:
## 🔍 Full audit + remediation (pushed as `7e872a5`)

Ran a multi-agent review (8 finder angles + adversarial verification) plus a dedicated security pass over the whole diff. **22 confirmed findings, all fixed in this PR.** Re-verified: backend 82 suites / 1321 tests, frontend 640 + clean build, Docker e2e 22/22 (regression) + 7/7 (fix-specific), `{meta.device:1, ts:1}` index confirmed in Mongo.

No High/Critical security vulnerabilities. Full write-up lives in `CODE_AUDIT_2026-07-10_CLIMATE.md` (kept untracked like the repo's other audit docs).

### High
- **Ingest atomicity** — persisting via `device.save()` read-modify-write meant a *legitimately retrying ESP32* could throw a `VersionError` **after** the readings were stored → firmware retry → duplicate readings; and parallel posts could bypass the daily cap. Now an atomic `updateOne` (`$inc` counter, `$set` caches).
- **Missing `meta.device` index** — the auto time-series index can't serve `meta.device`, so history charts and device deletion scanned across *all* tenants' readings. Added the secondary index.
- **Rejected readings mutated device state** — a bad value (e.g. `-127` from a disconnected probe) registered a phantom channel; a quota-trimmed reading still updated the "current" value and could fire a false alert. Now channels register / caches update only for fully-valid, in-quota readings.

### Medium
- `?range=constructor` prototype-chain key broke the readings query → `hasOwnProperty` guard.
- `climate` scope was mintable from the generic token API but could never bind to a device → excluded from user-mintable scopes.
- Device tokens rendered a raw `scope_climate` i18n key in the token list → string added (en+sv). *(A token-rotation endpoint so revoking a device token isn't terminal is noted as a follow-up.)*
- A dead probe's stale value alerted forever → channels older than the offline window are skipped.
- A device assigned to a soft-deleted cellar still displayed as armed → now ignored, matching the offline job.
- Saving thresholds snapshotted *all* defaults into user data → only the fields you actually set are stored, so unset ones keep following defaults.
- **Security (cross-tenant):** removing/downgrading a shared-cellar member didn't detach their climate devices → their token kept writing into your cellar and their alerts leaked its live name/thresholds. Now detached on removal and on editor→viewer downgrade.
- Frontend stale-render guards: `ClimateCard` reset + `fetchSeq` (no showing cellar A on cellar B), `ClimateHistoryModal` `fetchSeq` (no out-of-order range render).

### Low / cleanup / efficiency
Shared password-confirm limiter (was 3 separate stores multiplying the auth budget); shared `isValidId`; GDPR export now includes `lastRssi`; tooltip CSS shipped with the card instead of the lazy Statistics chunk (was rendering unstyled); shared `SERIES_COLORS`; offline cron bounded query; parallel cellar-fetch in ingest; card stops polling sensor-less cellars + pauses on hidden tab; lazy cellar-list fetch in Settings; localized last-seen units; dropped dead re-exports + an unused populate.

### Also
Removed 6 unrelated `vintage-availability` files that had been accidentally swept into the backend commit.

Two items unchanged **by design**: the per-channel `alertState` field (looked redundant but a cooldown-suppressed breach needs it), and the shipped defaults (730-day retention, 5 devices/user, 8–16 °C / 45–80 %RH, no plan gating) — still open for your call, all env/user-tunable.

Issue #698, opened 2026-07-11
FEATURE REQUEST: enable OpenAI SDK AI inference endpoints.


nathan-curtis:
Your application is hardcoding both a connection to voyageai and claude. requiring both keys. 

Please allow alternative OpenAI compatible Inference endpoints, Allowing selfhosters to provide selfhosted inference. 

Chat (aiChat.js) + label scan (labelScan.js): both instantiate new Anthropic({apiKey: ...}) directly from the Anthropic SDK
Embeddings: embedding.js hits https://api.voyageai.com/v1/embeddings as a hardcoded URL

Swapping to or adding support for the OpenAI SDK with baseURL (works against Ollama/LM Studio/vLLM's OpenAI-compat route) would be a moderate, well-scoped fork patch BUT open to the security conscious read: "you're not getting my Anthropic API key" crowd. :)

jagduvi1:
Thanks for the suggestion, we actually like this idea a lot.

Just so you know, it's a bit more work than only changing the URL. The code is built pretty tightly around the Anthropic API. The way messages, system prompt, streaming, image handling and token counts work is different from the OpenAI API, so it's not a one-line thing. And for the embeddings it's the same, the vector size is tied to the Qdrant collection, so we can't just point it somewhere else.

But we want to do it properly. The plan is to add a small provider layer inside the app, keep Anthropic and Voyage as the built-in ones, and make it open so other backends can be added. The main piece is an OpenAI-compatible option. Since Ollama, LM Studio, vLLM and so on all speak the OpenAI API, that one covers basically all the self-hosted cases. We'll also make the embedding vector size configurable, so a fresh self-hosted install can build its collection at whatever size its model uses.

It will be opt-in, cellarion.app itself keeps running on Anthropic and Voyage. We'll probably do the chat/LLM part first and the embeddings after. We've added this to our roadmap and will get it built in one of the coming releases. Thanks again for the great suggestion

nathan-curtis:
I'm aware of the work Iapped it out before I suggested it. Didn't want to suggest something not feasible...

But for me at least. Nit supporting a custom endpoint is a 100% blocker. (target local Ai on a Spark using vllm) 

If you support custom OAI style endpoints that's an unlock and I'm no longer blocked..

Btw I'm writing an integration for ZenOS-AI for your toolset in HA. 

...after the unblock.

jagduvi1:
[...]

PR #704, opened 2026-07-15
feat(ai): pluggable AI providers — OpenAI-compatible LLM + embedding endpoints for self-hosters

jagduvi1:
## Summary

Implements **both phases of #698** (LLM + embeddings — not marked as auto-closing so the issue can be closed after the reporter confirms on a real self-hosted endpoint): self-hosters can point every AI feature at any OpenAI-compatible endpoint (Ollama, LM Studio, vLLM, LiteLLM, OpenAI) instead of Anthropic + Voyage. **Opt-in via env** — without the new vars nothing changes; cellarion.app keeps running on Anthropic + Voyage exactly as before. The two switches are independent (e.g. local LLM + Voyage embeddings is valid).

## Phase 1 — LLM provider (`services/aiProvider.js`)

`getChatClient()` returns the Anthropic SDK client (default, unchanged incl. `maxRetries: 4` for imports) or an adapter that speaks `POST {OPENAI_BASE_URL}/chat/completions` while exposing the **same minimal surface the services already use**:

- `messages.create(params)` → `{ content: [{type:'text',text}], usage: {input_tokens, output_tokens} }`
- `messages.stream(params)` → events `text` / `finalMessage` / `error` / `abort` + `.abort()` (SSE parse, `stream_options.include_usage`)
- errors carry `err.status` + `retry-after`, so existing 429-retry / model-fallback / AI-budget refund logic works unchanged
- base64 image blocks (label scan) → `image_url` data-URLs; Anthropic-only `thinking` dropped; transparent 429/5xx retry with backoff

Model in openai mode comes from `AI_MODEL` / `AI_VISION_MODEL` env (admin Claude-model allowlist doesn't apply; configurable **prompts** still do). Covers cellar chat (incl. streaming), label scan, import lookup, text search, maturity/price/profile suggestions.

## Phase 2 — embedding provider (`services/embedding.js`)

`EMBEDDING_PROVIDER=openai` routes wine embeddings to `POST {EMBEDDING_BASE_URL}/embeddings`. Key safety property: **the Qdrant collection is sized from `getEmbeddingDimension()`** and every returned vector is validated against `EMBEDDING_DIMENSION` — a wrong value fails loudly instead of silently corrupting semantic search. Changing provider/model/dimension requires a **full embedding job** (existing rebuild path; documented in README). `WineEmbedding` bookkeeping records the model that actually embedded (`activeEmbeddingModel()`).

All `ANTHROPIC_API_KEY` / `VOYAGE_API_KEY` feature gates (import AI eligibility, enrichment job, chat RAG search, restock checker, superadmin health/status) are now provider-aware (`isConfigured()` / `isEmbeddingConfigured()`); the superadmin Services/AI tabs label the active provider.

## New env vars (all optional)

| Var | Meaning |
|---|---|
| `AI_PROVIDER` | `anthropic` (default) or `openai` |
| `OPENAI_BASE_URL` | `/v1` root of the LLM endpoint (required in openai mode) |
| `OPENAI_API_KEY` | bearer token (Ollama/LM Studio ignore it) |
| `AI_MODEL` | LLM model name (required in openai mode) |
| `AI_VISION_MODEL` | vision model for label scan (defaults to `AI_MODEL`) |
| `OPENAI_TIMEOUT_MS` | LLM per-request timeout, default 120000 |
| `EMBEDDING_PROVIDER` | `voyage` (default) or `openai` |
| `EMBEDDING_BASE_URL` | embedding `/v1` root (falls back to `OPENAI_BASE_URL`) |
| `EMBEDDING_API_KEY` | bearer token (falls back to `OPENAI_API_KEY`) |
| `EMBEDDING_MODEL` | embedding model (required in openai mode) |
| `EMBEDDING_DIMENSION` | the model's vector size (required in openai mode) |
| `EMBEDDING_TIMEOUT_MS` | embedding per-request timeout, default 30000 |

## docker-compose impact

- `docker-compose.yml`: 12 new pass-through env lines on the **backend** service only, all `${VAR:-}` (inert when unset). **No new services, no port/volume changes.**
- **Prod (`docker-compose.prod.yml`): no action needed** — cellarion.app stays on Anthropic + Voyage; mirror the vars only if openai mode is ever wanted there.

## Testing

- `aiProvider.test.js` (20 tests): provider selection/config, message translation (system/text/image), create (body shape, auth, vision-model pick, usage mapping, 429 retry + exhaustion, 400 no-retry, timeout→504), streaming (deltas + usage, split-chunk buffering, error status, abort suppression).
- `embedding.test.js` (+6 tests): provider helpers, endpoint/model/env fallbacks, input-order guarantee, **dimension-mismatch guard**, config errors, timeout shape.
- Full backend suite in Docker (node:20-alpine): **88/88 suites, 1431/1431 tests pass**. Frontend: **29/29 files, 647/647 tests pass**.
- ⏳ Full `docker-compose up` smoke test deferred until the parallel MCP session releases the Docker environment; recommended before merge. Real-endpoint validation (Ollama/vLLM) welcome — the reporter offered to test.


jagduvi1:
## Adversarial review pass (commit 02c9f37)

Ran an 8-angle review (line-by-line, removed-behavior, cross-file trace, reuse, simplification, efficiency, altitude, conventions) over the provider-layer diff. 7 confirmed bugs fixed + cleanups:

**Bugs fixed**
1. Stream flush — a final SSE `data:` line without a trailing newline dropped the last delta / usage frame (token accounting undercounted)
2. Date-format `Retry-After` → NaN sleep in the embedding retry loop (all retries burned instantly)
3. Unbounded `Retry-After` sleep — a hostile `3600` parked a chat/scan call for an hour; waits now capped at 32s
4. Retried 429/5xx response bodies never consumed → pooled undici connections held until GC
5. `ensureCollection` now loudly flags a dimension-mismatched collection (provider switch without the FULL re-embed) instead of letting every chat request fail with an opaque 500
6. openai-mode 'fallback' re-sent the identical request to the identical model and logged a fictitious second model — fallback now compares provider-resolved names and disables itself
7. A typo'd `AI_PROVIDER`/`EMBEDDING_PROVIDER` silently disabled all AI — now warns loudly once

**Cleanups** — one shared `utils/fetchRetry.js` replaces the two divergent retry loops; multi-MB label-scan bodies serialized once (not per retry); stream gained pre-first-token retry parity with the Anthropic SDK; per-row `isConfigured()` hoisted out of the import loop; `VOYAGE_DIMENSION` un-exported (sizing must go through `getEmbeddingDimension()`); SuperAdmin AI tab notice when model settings are env-governed; README privacy note for remote third-party endpoints.

**Consciously skipped:** removing `stream_options.include_usage` (all documented targets support it; removing loses token accounting), renaming the legacy `results.anthropic`/`voyageAI` API keys (compat), moving effective-model resolution into aiConfig (bigger refactor, noted for later), and the env-var name in the 503 message (it helps self-hosting operators more than it leaks).

Tests after fixes: backend **88/88 suites, 1434 tests** green in Docker; frontend **647** green.

jagduvi1:
## Follow-up: model resolution moved into aiConfig (commit 19202bf)

Closes the review's remaining altitude finding. `aiConfig.get()` now returns **provider-resolved** model fields — in openai mode every LLM model resolves to `AI_MODEL` (label scan to `AI_VISION_MODEL`), `embeddingModel` to `EMBEDDING_MODEL` — so event logs, WineEmbedding bookkeeping, and the fallback comparison are honest **by construction** instead of each call site remembering a translation wrapper. Default providers hit a fast path that returns the cache untouched: zero cost, zero change for cellarion.app.

New `aiConfig.getRaw()` serves the admin routes — GET /ai displays the stored Claude names (pickers keep working), and the five PATCH handlers merge-and-persist from raw (merging from the resolved view would have written env model names into SiteConfig).

The per-callsite wrappers are gone: `displayModel` deleted (replaced by `effectiveModels()`, consumed only by aiConfig), embeddingJob's three `activeEmbeddingModel` calls reverted to plain `cfg.embeddingModel`, aiChat's fallback check is a plain comparison again. New `aiConfig.test.js` pins the resolution contract.

Tests: backend **89/89 suites, 1437 tests** green in Docker.

jagduvi1:
## Final audit pass (commit 9f52199)

Second adversarial review over the whole branch (4 fresh angles incl. an end-to-end openai-mode trace and an aiConfig-accessor audit — the accessor audit came back fully clean). The theme of the surviving findings: **pre-existing code that only becomes wrong when the endpoint stops being Anthropic**. All fixed:

**Bugs fixed**
1. `extractFirstJsonObject` kept leading preambles ('Here is the JSON: {…}') and a lone quote in the preamble corrupted brace tracking — every parse failure was a **non-refundable AI-budget debit**, so a large import on a chatty local model could burn its whole budget on wines the model actually identified. `scanLabelFull` also switched from bare `JSON.parse` to the extractor.
2. `callClaudeJson`'s outer 429 retry trusted `Retry-After` blindly (NaN sleep on date-format, unbounded on hostile values) — the same hole fixed in `fetchWithRetry` last round, one layer up. Now numeric-only, capped at 60s.
3. Stream: `data: [DONE]` now terminates (keep-alive servers may never close the connection); in-stream `data: {\"error\":…}` frames emit **'error'** instead of delivering a truncated answer as success; first-token watchdog added (compat servers send 200 headers instantly, then stall loading the model — the header timeout never protected streaming).
4. **Data-loss guard**: a FULL embedding job with a broken embedding config used to DROP the collection and then fail to recreate it. `embeddingJob.start()` now refuses when unconfigured, `ensureCollection` refuses size ≤ 0, and incremental jobs fail fast with an actionable message on a dimension-mismatched collection.
5. `EMBEDDING_API_KEY` falls back to `OPENAI_API_KEY` **only when the base URL is shared too** — a dedicated embedding host is never sent the chat endpoint's bearer token.
6. `fetchWithRetry`: throwing `onRetry` hook can't abort the retry loop.

**Docs**: `host.docker.internal` is Docker-Desktop-only (Linux needs `extra_hosts: host-gateway` — the documented example would ENOTFOUND on the feature's primary audience); AI budgets tuned for Anthropic spend still meter free local endpoints (raise/disable in SuperAdmin → Rate limits).

**Noted, not changed**: `POST /api/chat` is the only live chat route — `chatStream`/the stream adapter currently ship without a production caller (pre-existing; the stream surface is fully unit-tested); stream consumers must attach an 'error' listener synchronously (documented in the class jsdoc).

Tests: backend **89/89 suites, 1442 tests** green in Docker (5 new: [DONE]-termination, error-frame, watchdog, key-isolation, preamble extraction).

PR #717, opened 2026-07-16
feat(mcp): export_cellar + get_account_export tools (Phase 1)

jagduvi1:
Closes the **last Phase-1 gap** in the MCP master plan (§3.3 / §3.18, Appendix A "Cellar export + GDPR account export"): data-portability exports reachable from the connected AI. Two tools + the redeem plumbing.

## Design: a link, not the bytes
An export is a **file**, not model context. The cellar payload caps at 50k documents and the account export spans ~35 collections — either can be hundreds of MB, and the account export carries PII (email, audit history). Inlining it would blow the caller's context window and hand the user's PII to the third-party AI. So each tool **mints a short-lived (1h) download link** and returns a compact summary; the user clicks it, the file streams straight to their device, and the model never sees it (plan principle #7).

## Tools
| Tool | Scope | Notes |
|---|---|---|
| `export_cellar` | `read` | Cellar data (JSON, or a ZIP with the user's own images once/week). Owner-only — a `cellar_id` the caller doesn't own → `not_found`. |
| `get_account_export` | **`write`** | Full GDPR account export. Gated **above** `read` on purpose. |

**Why `get_account_export` is write-scoped (deviation from the plan's casual `[read]`):** the account bundle is a PII superset the `read` tier deliberately cannot reach today — `whoami` returns only `{id}`, `/me` stays 403 for tokens, and the `ApiToken` model's own contract says *"a leaked read-only token cannot touch exports"*. So a cautious read-only connection can never exfiltrate the account PII bundle; only a token the user granted full write trust surfaces it. Flagging for review.

## Plumbing
- **`services/exportLinks.js`** owns the `celx_` credential lifecycle (`celx_` + 32 random bytes, SHA-256 at rest, 1h TTL) and the redeem/stream path.
- The **weekly image-export allowance + ZIP streaming** are extracted from `routes/users.js` into `services/cellarExport.js`, so the web full-export route and the MCP redeem route run **one** implementation — MCP is not a bypass of the weekly cap. The account export adds a **once-per-day** throttle (`User.lastAccountExportAt`) **on the MCP path only**; the web route (`GET /api/users/me/export`) is unchanged.
- **`GET /api/mcp/export/:token`** is deliberately **unauthenticated** — the opaque token *is* the credential, so the link works in any browser and mints no login. `Cache-Control: no-store`.
- `ExportLink` registered in the GDPR erasure/export registry (purged on account deletion; nothing to export — an ephemeral credential isn't user data).
- Server `instructions` teach the AI the link semantics + the web-only account-root actions; corrected a stale "cannot create racks/cellars" line (create tools shipped in #715).

## Testing
- `services/exportLinks.test.js` — token hashing, expiry, throttle claim/refund, redeem branches (**real execution** vs mocked models, per the audit lesson).
- `mcp/exportTools.test.js` — scope split, owner-only guard, link envelope.
- **Full backend suite green: 103 suites / 1616 tests** (alpine).
- **Live e2e vs the local stack** (real Mongo/auth/redeem): read token can't see or call `get_account_export`; unauthenticated redeem returns real cellar + account JSON; 2nd account export in a day → 429; bogus/malformed token → 404.

## Compose impact
**None** — in-process, no new service/port/env/volume.

⚠️ No new user-facing strings requiring sv translation (tool descriptions + AI-facing error messages are English by design, like the rest of the MCP surface).


jagduvi1:
Adversarial audit ran (security + correctness over the whole diff): **0 High / 0 Med**. The two properties that matter — the structural read/write scope split and the owner-only data scoping — both hold. Follow-up commit `7450ed5` addresses the Low findings worth fixing:

- **Audit-log the download** (not just the mint) — redemption is where the PII actually leaves us; now recorded as `user.export_download` with the downloader IP.
- **Refund the weekly ZIP allowance on mid-stream disconnect** so a still-valid link stays retryable (a broken pipe previously left the allowance spent → re-click 429'd).
- **Explicit kind dispatch** (unknown → 400) and **unified 404 wording** (no existence oracle).

Verified live again after the fixes: 17/17 e2e checks pass, and the `user.export_download` audit rows land with the right kind/resource. Full backend suite 103 suites / 1619 green.

PR #725, opened 2026-07-17
feat(mcp): attach_bottle_image — add a label/bottle photo over MCP

jagduvi1:
## What this is

Closes the §3.18 parity gap a real user hit in live testing: they added 5 bottles by talking to their AI, then asked it to attach the label photo — and no tool could. This adds `attach_bottle_image` and, in doing so, extracts the image pipeline into a shared service so the REST upload and MCP run one implementation.

**⚠️ STACKED: #721 → #722 → #723 → #724 → this.** Merge in order.

## The tool

`attach_bottle_image` [write] — two input modes for a caller with no file-upload channel:
- **image_url** (https) — a product/label image found on the web (e.g. a retailer's wine page). Fetched **server-side through a full SSRF guard**.
- **image_base64** — bytes the user showed their AI directly, capped ~1.1MB to fit the JSON body.

Confirm-before-attaching in the instructions; **reversible** via `undo_last` / the activity timeline (deletes the BottleImage row + files, scoped to the caller's own non-registry-assigned image). Background removal (rembg) runs automatically, same as the web upload.

## The SSRF guard (`utils/safeImageFetch.js`) — the crown jewel

The backend sits on a Docker network with internal services and a VM with a cloud-metadata endpoint, so a naive `fetch(url)` is a server-side request forgery hole. The guard, in order:
- **https + default port only**; no embedded credentials.
- **DNS resolved first, then the socket PINNED** to the validated address via a custom `lookup` — a DNS-rebind flip between check and connect can't redirect it. SNI/Host stay the real hostname so legit CDNs still resolve.
- **Every A/AAAA record must be public unicast**; a mixed set is refused.
- **Redirects followed manually (≤3), each hop re-validated from scratch** — a public host 302ing to `http://169.254.169.254` is refused.
- 10s timeout, response **streamed with a 10MB abort cap** (not buffer-then-check).
- Address classifier blocks: 0/8, 10/8, 127/8, 100.64/10 (CGNAT), 169.254/16 (link-local + metadata), 172.16/12, 192.168/16, 198.18/15, 224/4+, 255.255.255.255; v6 ::, ::1, fe80::/10, fc00::/7, 64:ff9b::/96 (NAT64), and **all embedded-v4 IPv6 literals**.

**Self-audit caught a real bypass and this PR fixes it**: `https://[::ffff:169.254.169.254]/` is normalized by Node to the *hex* form `::ffff:a9fe:a9fe`, which a dotted-decimal-only v4-mapped check missed — it would have reached the metadata endpoint. Now all embedded-v4 v6 literals are blanket-refused (no legit image host is one). Decimal/hex/octal IP encodings (`2130706433`, `0x7f000001`, `017700000001`) are already defused because Node's URL parser normalizes them to `127.0.0.1` → caught by the literal-IP check. **SVG is not in the supported formats** (jpeg/png/webp only via sharp), so the SVG-XSS class can't land.

## One implementation for UI and MCP

`services/imageOps.ingestBottleImage` is extracted from the REST `/upload` route and now backs **both**: `sanitizeImageBuffer` (decode + validate + strip EXIF/GPS, **fail closed**), the per-bottle cap, the sniffed-format extension (never a caller claim), persistence under a random UUID name, and the rembg hand-off. The route keeps multer + authz and delegates the bytes. Its old **fail-open** header dimension check is gone — the shared pipeline's pixel cap fails closed (a strict improvement). A disk write failure returns a clean error instead of crashing the tool (→ `rate_limited`, retryable).

## Verification

- **SSRF guard suite (37 tests)**: every private/reserved range for v4 and v6, all v4-mapped textual forms incl. the hex-normalized exploit, scheme/port/credential/redirect rules.
- **imageOps suite**: fail-closed sanitisation, per-bottle cap, sniffed extension, bg hand-off.
- **Tool suite**: URL/base64 XOR, access re-check, SSRF error surfaced, ledger row, idempotency replay, undo scoping.
- **Smoke + live Docker e2e ALL OK**: base64 attach → **visible via `GET /api/images/bottle` (the same pipeline the web upload feeds)** → `undo_last` removes it → a cloud-metadata URL is refused.
- **Full backend suite: 1771 tests green** (node:20-alpine).
- Adversarial audit running; any findings fold into this branch.

## Body-limit / compose

`/api/mcp` raised 10kb → 2mb for the base64 mode (registered **after** `/api/mcp/public`, which stays 10kb — the anonymous surface has no image tool). No new service/env/dependency. `McpActionLog.action` gains `attach_image` (additive).

### DIFF
--- a/backend/src/app.js
+++ b/backend/src/app.js
@@ -107,6 +107,13 @@ app.use('/api/bottles/import/sessions', express.json({ limit: '5mb' }));
 app.use('/api/bottles/import', express.json({ limit: '8mb' }));
 app.use('/api/blog/admin/posts', express.json({ limit: '2mb' }));
 app.use('/api/wine-lists', express.json({ limit: '1mb' }));
+// MCP: the anonymous public surface stays tiny (no image tool) — registered
+// BEFORE the personal /api/mcp rule so /api/mcp/public matches it first (the
+// first express.json to parse wins). The personal endpoint allows a larger
+// body for attach_bottle_image's base64 mode (MAX_BASE64_CHARS ≈ 1.5MB + JSON
+// overhead); every personal MCP caller is authenticated and rate-limited.
+app.use('/api/mcp/public', express.json({ limit: '10kb' }));
+app.use('/api/mcp', express.json({ limit: '2mb' }));
 app.use(express.json({ limit: '10kb' }));
 const corsOrigin = process.env.FRONTEND_URL || (process.env.NODE_ENV === 'production' ? false : 'http://localhost:3000');
 if (process.env.NODE_ENV === 'production' && !process.env.FRONTEND_URL) {
--- a/backend/src/mcp/consumeTools.test.js
+++ b/backend/src/mcp/consumeTools.test.js
@@ -285,7 +285,7 @@ describe('undo_last walk-backward + scope gating (e2e-caught regression)', () =>
     await tool('undo_last').handler({}, { ...CTX, scopes: ['consume', 'write'] });
     q = McpActionLog.findOne.mock.calls[0][0];
     expect(q.action.$in).toEqual(['consume', 'restore', 'add', 'update', 'bulk_add', 'somm_maturity', 'somm_price',
-      'cellar_create', 'rack_create', 'place', 'unplace', 'move', 'arrange', 'tasting_note']);
+      'cellar_create', 'rack_create', 'place', 'unplace', 'move', 'arrange', 'tasting_note', 'attach_image']);
   });
 
   test('the reverse row is marked viaUndo and the undone row frees its idempotency key', async () => {
--- a/backend/src/mcp/eval/cases.js
+++ b/backend/src/mcp/eval/cases.js
@@ -146,6 +146,12 @@ const CASES = [
     prompt: 'Do you have a guide on how to store wine properly long-term?',
     expect: { tool: 'list_guides' },
   },
+  {
+    id: 'attach-photo',
+    prompt: 'Here is the label photo for that Barolo — https://example.com/barolo-label.jpg — put it on the bottle.',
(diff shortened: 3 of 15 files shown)


jagduvi1:
**Adversarial security audit complete — GO at HEAD.**

An independent SSRF-focused audit converged on exactly the bypass I'd self-caught mid-build: `https://[::ffff:169.254.169.254]/` normalizes to the hex form `::ffff:a9fe:a9fe`, which the old dotted-decimal-only v4-mapped check missed. Verified fixed by `84a2b4d` (blanket-refuse all embedded-v4 v6 literals) end-to-end. It was the only real gap — capped at Medium because the guard pins port 443 and cloud metadata serves HTTP:80, so exfil was blocked even at `48fa638`; residual was a TLS-on-443 internal scan (nothing in the compose speaks that).

Every other vector came back clean: DNS pin / rebinding / TOCTOU, per-hop redirect re-validation, decimal/hex/octal IP encodings (defused by URL normalization), decompression bombs, SVG-XSS (not a supported format), base64/body-limit ordering, REST-route parity (the removed checks were weaker + fail-open), undo scoping (`uploadedBy`+`assignedToWine:false`), anonymous double-gating, and audit-log URL leakage.

**Two Lows:**
- **L1 — fixed (`6319b82`):** the "10s total timeout" was a socket *idle* timer (a slowloris origin dribbling bytes under the 10MB cap evaded it). Added a real wall-clock deadline over the whole request via a single-settle guard. The claimed guarantee is now real.
- **L2 — deferred fast-follow:** `replay()` reads before the ledger row is written, so two concurrent same-`idempotency_key` calls both persist → one image with no ledger row (un-undoable). Caller's-own-bottle nuisance, bounded by the 20/bottle cap + mutation budget; matches the pre-existing MCP idempotency+undo caveat. Fix = pre-claim the key before acting. Recommend before the write surface sees real load.

SSRF guard suite now 44 tests.

PR #819, opened 2026-07-24
Add german translation

susertux:
german translation, done with perplexity.ai and mistral.ai vibe


jagduvi1:
Hi @susertux — thank you for this! A German translation is genuinely wanted, and we'd love your help getting it shipped.

Cellarion's translations are managed through Weblate rather than direct PRs, so I'm closing this one per our workflow (see [TRANSLATING.md](https://github.com/jagduvi1/Cellarion/blob/main/TRANSLATING.md)): direct edits to `translation.json` conflict with Weblate's two-way sync, and a new language goes live once it reaches ~90% translated (untranslated strings fall back to English until then, so it degrades gracefully while you work).

Your work won't be lost — we'll import this file into Weblate **as suggestions**, so every string just needs a review/accept there rather than retranslating from scratch. If you'd like to drive German, it's here: **https://hosted.weblate.org/projects/cellarion/** (no Git knowledge needed). Do check the glossary for the wine-domain terms (drink window, maturity, rack…) — that's where AI drafts usually need the most correction.

Thanks again, and hope to see you in Weblate! 🍷

PR #839, opened 2026-07-26
fix(registry): stripProducerSuffix must not leave a stranded connective

jagduvi1:
## The tap behind the ticket

#837 ships the scan that FINDS dangling names (`dangling-name-tail.v1`); this PR stops new ones being minted. "La Viña de Madaras" (producer "Madaras") was stripped to **"La Viña de"** at create time — `stripProducerSuffix` is exact-string with only a something-remains guard, and `canonicalizeWineName` loops it on every registry write surface (findOrCreateWine, admin create, wine-request approval). 10 such rows existed on prod when the ticket was filed.

## The guard

After the strip, if the remainder's last normalized token is in `DANGLING_TAIL_WORDS` (the same list the scan rule uses — one list, one behavior), refuse the strip and keep the full phrase. Rationale: when the producer token is grammatically part of the name, the full phrase **is** the name — the "Les Forts de Latour" family the suffix stripper already respects by staying exact-string. Keeping a producer-embedded name (reviewable) beats writing a broken one (invisible).

Only the **tail** token is inspected — "Fiano di Avellino Mastroberardino" still strips to "Fiano di Avellino". Prefix strips untouched (a leading-edge strip can't create this shape).

## Base branch note

**Based on `feat/registry-name-check-verification` (#837)** — it uses the `DANGLING_TAIL_WORDS` set that lands there. Merge #837 first; this then retargets to `main` (I'll rebase it so the diff is clean).

## Tests

6 suites / 136 tests green in `node:20-alpine`, including the two real prod rows ("La Viña de Madaras", "Re di Renieri"), von/du connectives, diacritics/case immunity, the interior-connective happy path, and `canonicalizeWineName` settling on the original name. The #837 snapshot suite still passes — no `detect()` source changed (a shared-helper behavior change is the documented residual there, called out in `nameChecks.js`).

## Docker-compose impact

**None.**

### DIFF
--- a/backend/src/utils/producerPrefix.js
+++ b/backend/src/utils/producerPrefix.js
@@ -74,7 +74,17 @@ function stripProducerSuffix(name, producer) {
   if (!/[\s\-–—]$/.test(rest)) return null;
 
   const remainder = rest.replace(/[\s\-–—]+$/, '').trim();
-  return remainder.length > 0 ? remainder : null;
+  if (remainder.length === 0) return null;
+  // Never leave a stranded connective at the tail: "La Viña de Madaras" with
+  // producer "Madaras" must NOT strip to "La Viña de" — exactly how prod's 10
+  // dangling-name rows were minted (support ticket 2026-07-26). When the
+  // producer token is grammatically part of the name, the full phrase IS the
+  // name (same family as "Les Forts de Latour"); keeping it beats writing a
+  // broken one. canonicalizeWineName loops this function on every registry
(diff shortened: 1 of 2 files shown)


jagduvi1:
Superseded — this PR's base branch (`feat/registry-name-check-verification`) was deleted when #837 squash-merged, which auto-closed this and made the base unchangeable. The branch has been rebased onto `main` (clean 2-file diff) and re-opened as a new PR against `main`. No content change.

PR #854, opened 2026-07-29
feat(somm): let a curator remove an unreleased vintage from the maturity queue

jagduvi1:
**Stacked on #853** — base is `feat/somm-curate-wine-profile`, so the diff here is only this feature. ⚠️ When #853 squash-merges, this needs `git rebase --onto main feat/somm-curate-wine-profile` (GitHub retargets the base but the squash makes #853's commits reappear in the diff until then). CI also only runs once retargeted — `ci.yml` triggers on PRs against `main` only — so the verification below is local.

## The problem

A queue entry whose vintage the wine does not have cannot be curated at all. A user types a year the wine was never released in — most often while testing, sometimes a purchase or disgorgement year mistaken for a vintage — and the pair sits in the queue forever. Both existing exits are wrong for it: **Mark reviewed** invents a drink window on data shared with every owner of that wine, and a discard did not exist.

## The design (Johan's, and simpler than my first cut)

**Remove = delete the row.** No new status, no new fields, no schedule. The comeback mechanism already exists: `ensurePendingVintageProfile` re-seeds the pair on **every bottle add**, so the moment anyone adds a bottle of that wine+vintage again — for instance once the vintage is actually released — it re-enters the queue like any other wine. A bottle add is exactly the signal that the vintage now exists in the real world; the date-based deferral I built first (see the branch history) was guessing at the same event with a calendar. Scrapping it also deleted ~900 lines: no model migration, no third status leaking into stats/filters/exports, no new GDPR fields, no extra tab.

| | |
|---|---|
| REST | `DELETE /api/somm/maturity/:id` — 409 on a reviewed row (reset first; its window is live data users already see) |
| MCP | `remove_from_maturity_queue` (+ optional `reason` for the audit trail); reversible via `undo_last` |
| UI | a "Remove from queue" action on every pending card, behind a confirm that explains the comeback rule |

**Undo re-seeds** the pending stub via a direct upsert — deliberately not the swallowing `ensurePendingVintageProfile` helper, because an undo must fail loudly so the ledger row can be unclaimed and retried. Upsert rather than insert means an undo racing a genuine bottle add still lands on "it's in the queue again" instead of a unique-index error.

## Also in this PR: three defects fixed in #853's feature

Found while wiring into the same ledger machinery, all in `set_wine_profile`, all meaning its advertised undo could not have worked:

1. Its action string `somm_wine_profile` was **not in `McpActionLog`'s enum** — the ledger write failed validation.
2. It was **not in `WRITE_REVERSIBLE`** — `undo_last` could never select the row.
3. Its revert branch **never claimed the row**, unlike all other branches — undo would repeat forever and a concurrent twin could double-restore.

Plus one GDPR gap from the audit that still applies: `aiProfile.verifiedBy` (#853's curator ref) was missing from the erasure cascade in `userDataRegistry.js` — its sibling `createdBy` has been reassigned to the `[deleted]` sentinel for ages, and now `verifiedBy` is too. Note the registry's completeness test can't catch this class (it checks the *model* is registered, not every User-ref *field*); worth tightening separately.

## Audit trail

The branch was audited (3 independent passes: security / correctness / code review) while it still carried the deferral design — see the earlier comment. Findings specific to the deferral died with it; the ones above are what survived the redesign.

## Testing

- MCP tool tests: delete + ledger `prev` shape, reviewed-row refusal, undo-upsert + claim assertion; `set_wine_profile` behaviour + undo + double-undo-refused
- **963 backend tests** across all 70 `src/mcp/`, `src/services/`, `src/routes/somm/` suites; **801 frontend**; all green
- English strings only, no other locale touched

## Not in scope

Stopping the bad vintage getting in at all — the open Annapurna ticket's AddBottle vintage-path question is a separate fix, and this is what a curator needs regardless.


jagduvi1:
## Audit pass — 3 independent reviews of this diff (security / correctness / code review)

Ran after opening the PR, scoped to `git diff origin/main...HEAD` so it covers **both** this feature and #853. Seven findings survived my own verification; all seven are fixed in `1212c3e`. I read the code for each before acting rather than taking the reports at face value — two further claims did not survive and are listed at the bottom.

**The one that mattered — GDPR erasure cascade.** Both new curator references were missing from `services/userDataRegistry.js`: `WineVintageProfile.deferredBy` (mine) and `WineDefinition.aiProfile.verifiedBy` (#853's). Deleting a sommelier's account left both dangling at a user document that no longer exists, on shared registry data, indefinitely. Their siblings `setBy` and `createdBy` have been handled for ages — the precedent was right there in the same file. Note the existing completeness test at `userDataRegistry.test.js:27` **cannot** catch this class: it asserts every model carrying a `ref: 'User'` is *registered*, not that every User-ref *field* on it is purged. It stayed green throughout. Worth deciding separately whether that test should be tightened.

**State-machine hole.** MCP `set_vintage_maturity` set `status: 'reviewed'` without clearing the deferral, unlike the REST route — so reviewing a deferred row through the MCP surface left a reviewed row still carrying a return date. Fixed, plus the deferral now rides in the undo snapshot and the undo restores it: without that, undoing a review of a deferred row would silently drop the deferral rather than putting the row back as it was.

**Admin stat drifted.** `globalStatsService`'s "Pending profiles" counted `status: 'pending'` literally, so once sommeliers start deferring, it under-reports the very backlog the card's warn-above-50 threshold exists to flag. It now uses `buildQueueFilter('pending')`. `profilesTotal` no longer equals reviewed + pending — the gap is the deferrals still being held, noted in a comment. A separate `profilesDeferred` stat + label would be the fuller fix; say the word.

**Smaller four:** `GET /lookup` (reachable by *any* authenticated user, unlike the somm-gated queue) no longer returns the curator-facing deferral internals; an unused `DEFERRAL_FIELDS` export removed; the "pending first" sort claim made accurate now that three statuses sort lexicographically; and `set_wine_profile` gained the MCP behaviour + undo tests it never had — including one that pins the missing-claim fix by asserting a second undo of the same row is refused.

**Did not survive verification:** the pagination skip/duplicate concern is unreachable in practice (`MIN_DEFER_YEARS = 1` means no deferral boundary can fall inside a live pagination session), and the `parseDeferUntil` `instanceof Date` branch being test-only is true but harmless defensive coding.

### CI note
This PR targets `feat/somm-curate-wine-profile`, and `.github/workflows/ci.yml` triggers only on `pull_request: branches: [main]` — so **no checks will run here until it is retargeted**. Verified locally instead: **943 backend tests** across all 70 `src/mcp/`, `src/services/` and `src/routes/somm/` suites, and **801 frontend** tests, all green after the fixes.

jagduvi1:
Reworked per Johan's direction: the date-based deferral is replaced by plain removal — delete the row, and the existing bottle-add re-seed brings it back when anyone adds that wine+vintage again. Net −900 lines vs the first design; PR body rewritten to match. The #853 ledger fixes and the verifiedBy GDPR fix survive unchanged.

PR #858, opened 2026-07-29
feat(registry): the appellation system — membership list, write-time adoption, review queue

jagduvi1:
Fourth PR of the registry-quality campaign (`REGISTRY_STRATEGY_2026-07-29.md`, item **R2** — the structural centerpiece).

## The problem

Appellations had *nothing*: a free string on the wine (1,077 distinct strings across ~4k wines), an `Appellation` collection wines never referenced, and no membership list — so `appellation_wrong` was the 2026-07-26 audit's largest category (231 rows), and rules had to parse grammar where they should test membership (the "Clairette de Die" false positive, #850: "Die" read as a German article instead of the Drôme town).

## The design: the Grape treatment, not the Country treatment

Countries are a closed set — unknown is rejected. Appellations are genuinely open-ended (new AVAs and DOCs appear; "Vin de France" and "Wine of Origin Western Cape" look wrong and aren't), so an unknown value is **reviewed, never rejected** — a user adding a bottle is never blocked by taxonomy. Three parts:

**1. Write-time adoption** (`services/appellationResolve.js`): the tier-stripped appellation is looked up by normalized name OR synonym; a curated match adopts the canonical display spelling **before key generation**, so "Chateauneuf du Pape" produces the same `normalizedKey` as "Châteauneuf-du-Pape" and the exact-match stage collapses them instead of minting a sibling. Cross-country name clashes (the per-country unique index allows them) adopt only when the curated spellings agree. Non-fatal by construction. `Appellation` gains `synonyms`/`normalizedSynonyms` with the same lockstep save-hook as `Grape`.

**2. Seeding** (`scripts/seed-appellations-from-wines.js`, dry-run default): promotes every normalized appellation string carried by **≥ 3 wines** (the same floor the public taxonomy pages use) into a curated doc — majority display spelling (ties → oldest, the same rule as producer unification), majority country, region only when one covers at least half the group's wines in that country. The long tail below the floor deliberately stays free text: the point is to canonize the clearly established, not to bless every typo. ⚠️ Before running on prod, cross-check the output against `REGISTRY_AUDIT_2026-07-26.md` §4's leave-alone list.

**3. Review queue** (`GET /api/admin/taxonomy/appellations/unmatched` + the AdminTaxonomy "Review unmatched" modal): every string no doc covers, with majority spelling, wine count, and suggested country/region. One click promotes it through the normal create endpoint — the name is editable first, because the majority spelling of a rare appellation can itself be the typo. A suggested region from a different country than the suggested one is dropped (schema would reject it). The appellation create/edit form gains a synonyms field.

Together with #856's weekly cron this closes the loop: new variants can't form (adoption), the established ones are curated (seed), the tail is visible (queue), and future rules can *test membership instead of parsing grammar* — the "de Die" class of false positive dies here.

## Testing

- New `appellationResolve.test.js` (6) — adoption by name and by synonym, verbatim pass-through, cross-country agree/disagree, empty short-circuit, non-fatal failure
- `findOrCreateWine.test.js` extended — synonym input produces the canonical `normalizedKey` (the collapse that matters), un-curated stored verbatim
- Full sweep: **1004 backend tests / 78 suites**, **801 frontend**, all green

## Prod notes

No compose/env impact. After deploy: run the seed script (dry-run first, review the list, then `--apply`). The write-time resolver is inert until docs exist, so the deploy order is safe either way.

### DIFF
--- a/backend/src/models/Appellation.js
+++ b/backend/src/models/Appellation.js
@@ -1,4 +1,5 @@
 const mongoose = require('mongoose');
+const { normalizeString } = require('../utils/normalize');
 
 const appellationSchema = new mongoose.Schema({
   name: {
@@ -23,6 +24,19 @@ const appellationSchema = new mongoose.Schema({
     default: null,
     index: true
   },
+  // Alternative spellings that resolve to this appellation at write time
+  // ("Chateauneuf du Pape" → "Châteauneuf-du-Pape") — same design as
+  // Region/Grape synonyms, so a variant can't quietly become a second string
+  // on wines. normalizedSynonyms is what lookups match against.
+  synonyms: {
+    type: [String],
+    default: undefined
+  },
+  normalizedSynonyms: {
+    type: [String],
+    default: undefined,
+    index: true
+  },
   createdBy: {
     type: mongoose.Schema.Types.ObjectId,
     ref: 'User',
@@ -36,5 +50,15 @@ const appellationSchema = new mongoose.Schema({
 
 // No two appellations with the same name in the same country
 appellationSchema.index({ country: 1, normalizedName: 1 }, { unique: true });
+
+// Keep normalizedSynonyms in lockstep with synonyms — same hook as Grape, so
+// every save-based admin edit maintains what the write-time lookup matches.
+appellationSchema.pre('save', function(next) {
+  if (this.isModified('synonyms') || this.isNew) {
+    const normalized = (this.synonyms || []).map(s => normalizeString(s)).filter(Boolean);
+    this.normalizedSynonyms = normalized.length ? normalized : undefined;
+  }
+  next();
+});
 
 module.exports = mongoose.model('Appellation', appellationSchema);
--- a/backend/src/models/Appellation.js
+++ b/backend/src/models/Appellation.js
@@ -1,5 +1,5 @@
 const mongoose = require('mongoose');
-const { normalizeString } = require('../utils/normalize');
+const { normalizeAppellationKey } = require('../utils/normalize');
 
 const appellationSchema = new mongoose.Schema({
   name: {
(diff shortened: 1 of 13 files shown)


jagduvi1:
## Tested against a real prod copy — 3 fixes pushed

Per Johan's direction, restored last night's prod backup (4,650 wines / 7,877 bottles / 228 users) into local Docker running this branch, and drove the real add-bottle API against it. **Final result: 14/14 scenarios green** — exact match on a prod wine, bottle add (year + NV, with correct maturity-queue seeding and no downgrade of a reviewed profile), accent-variant producer adoption (#855 verified on the actual Ribeauvillé rows), soft-zone hold for a 1-character typo (measured score 0.948 → "did you mean", not a silent create), appellation adoption, and all four write gates (junk country, producer-is-a-place, comma region, junk grape).

The prod copy caught **three defects no unit suite could see**, all fixed in the new commit:

1. **`normalizeString` deletes hyphens** → the seeded "Châteauneuf-du-Pape" doc and the typed "Chateauneuf du Pape" produced different keys and the resolver never matched its own taxonomy. New `normalizeAppellationKey` folds hyphens to spaces, applied consistently across all seven appellation touchpoints. Deliberately appellation-scoped: `normalizedKey`/`canonicalKey` semantics untouched.
2. **Pre-guard tier suffixes** — stored strings like "Ribera del Duero DO" (written before the tier-strip guard existed) were about to become tier-suffixed taxonomy docs; seed + unmatched queue now fold them into their clean form first.
3. **Double-space spellings can win the majority vote** — "Wrights  Estate"×2 vs "Wrights Estate"×1. Mint now collapses internal whitespace; the unify script groups candidates collapsed and still rewrites rows whose stored string differs.

Real-data dry-runs, for when this deploys:
- **Seed**: 332 appellations at the ≥3-wine floor (from 1,064 distinct unmatched strings; 739 below floor stay in the review queue). "Vin de France" correctly created as-is per the audit's leave-alone list.
- **Unify**: 33 split groups / 36 rows, **1 key-drift safety skip** — the safety proof caught a non-trivial variant exactly as designed.
- **Health job** baseline on prod data: 5 name flags, 2 collision sets, 86 low-confidence, 34 producer splits, **0 missing invariants**, 0 open reports.

⚠️ **One operational finding for the restore runbook** (not this PR's code): after a mongo restore, Meilisearch keeps its old index — boot-time sync skips when the index looks populated, and until a force-resync the dedup net silently degrades (that's what made the typo test create a duplicate in early runs). The in-process resync can be triggered via an admin taxonomy rename; a dedicated admin "force reindex" endpoint is a candidate for the next PR.

1165 backend tests / 88 suites green after the fixes.

PR #859, opened 2026-07-29
feat(registry): region review queue, search force-reindex, label-abbreviation matching

jagduvi1:
Fifth PR of the registry campaign (`REGISTRY_STRATEGY_2026-07-29.md` **R3** + the ops items the prod-copy testing surfaced).

## R3 — regions minted by users are now witnessed

`findOrCreateRegion` auto-creates a Region doc for any comma-free, non-country string — that silence is why the collection needed cleanup scripts twice in July. A user-write mint is now born `pendingReview: true`, shown in AdminTaxonomy with a "New — unreviewed" badge and a one-click **Approve** (its own audited verb — editing a region to fix a typo must not silently count as "a human vouched for this"). The user is never blocked; the mint is just visible. Admin creations and the curated LWIN import stay born-reviewed. Existing regions are untouched (flag defaults false) — no migration.

## Force-reindex — the restore runbook's missing button

Found live during the prod-copy testing: after a mongo restore, the boot-time Meilisearch sync **skips** ("already populated"), and a stale index **silently disables the fuzzy dedup net** — near-duplicate adds mint duplicates with no error anywhere. The only in-process workaround was renaming a country to itself. Now: `POST /api/admin/search/reindex` (awaited to completion — the response means *done*) + a "Reindex search" button in AdminWines with duration toast.

## Label-abbreviation matching — from the 15-photo scan bench

Every German GG label measured under the 0.75 scan-match floor purely on the "GG" vs "Großes Gewächs" token. The name axis now expands label abbreviations (`gg`, `1er`) on **both sides at comparison time only** — stored strings and keys never change. Two subtleties the tests pin: the expansion targets the *actual* fold (`normalizeString` deletes ß, so "Großes Gewächs" → `groes gewachs` — an expansion to the intuitive `grosses gewachs` would never match), and word-boundary anchoring so "Eggenburg" stays intact. Measured on the bench's real cases: Kuhn GG **<0.75 → 0.902**, Barraud 1er 0.81 → 0.908 — both now clear the match floor.

**Considered and rejected**: a name-equals-known-appellation name-check rule (from the bench's Sixtine finding). Too many wines are *legitimately* named exactly their appellation (Barolo, Chablis, Sancerre base bottlings) — the rule would flag half of Piedmont. The two malformed prod rows the bench found (Sixtine name=appellation, Thébaud/Thivin producer/name swap) are individual data fixes for the queue instead.

## Testing

- 3 new `wineMatching` tests (GG expansion, 1er expansion, no mid-word firing); region-mint test updated for the flag
- **1077 backend tests / 87 suites** and **801 frontend** green
- No compose/env impact; no migration

Remaining campaign item: R7 (structured report corrections + public-page report link) — next PR.


jagduvi1:
**Extended with MCP parity** (per Johan's question — most curation happens from chat): four new admin-only MCP tools mirror the two review queues.

- `list_region_review_queue` / `approve_region` — the pending-region queue with wine counts; approve is idempotent and audits the **same action string** as the REST route (plus `via:'mcp'`). Merge is deliberately *not* exposed over MCP — it rewrites every referencing wine and stays a web-admin decision.
- `list_unmatched_appellations` / `promote_appellation` — the appellation queue; promotion creates the curated doc (hyphen-folded key + synonyms), after which new wines adopt the canonical spelling automatically.

The queue logic moved into `services/taxonomyReview.js`, shared by REST and MCP so the surfaces can't drift — same pattern as `wineProfileOps`. All four are structurally admin-only (unregistered for non-admin connections, like the somm tools), and follow `admin_add_registry_wine`'s no-undo-ledger stance: audited, not personally undoable. MCP server instructions updated (the drift guard enforces this).

Suites after the extension: **1,383 backend tests / 108 suites** green (5 new MCP tests incl. structural invisibility, idempotent approve, hyphen-folded promotion, and the unique-conflict path), frontend unchanged.

PR #860, opened 2026-07-29
feat(registry): structured report corrections + public wine page report link

jagduvi1:
**The final PR of the registry campaign** (`REGISTRY_STRATEGY_2026-07-29.md`, item **R7**). With this, all seven roadmap items are built.

## The two gaps

The report flow worked end to end (modal → admin queue → Cellar Cred reward) but had the two flaws the strategy called out:

1. **Resolution wrote nothing back.** The admin read "the producer is misspelled", clicked Resolve, then re-typed the fix in Admin → Wines — or, realistically, didn't.
2. **The link only existed on the bottle page**, where only *owners* are. Wrong shared data is most visible to non-owners browsing the public wine page.

## What this adds

**Structured corrections.** When the reason is "wrong info", the reporter can optionally pick a field (name / producer / appellation / type — the one-click-appliable string fields; region/country are refs and stay free-text in details) and type the correct value, with the current value shown as the placeholder. Validated both-or-neither server-side; type is validated against the wine-type enum at submission so a junk suggestion can't even enter the queue.

**Apply & Resolve.** The admin queue shows the suggestion as `field: current → proposed`, and a report that carries one gets an **Apply & Resolve** button. Applying goes through `.save()` so the model hooks run (canonicalKey recompute + `verifiedChecks` invalidation — a renamed wine must re-earn its clearances), with `normalizedKey` recomputed exactly as the admin edit route does it. The subtle case is pinned by a test: if applying the correction would make the wine **identical to an existing registry wine**, that's a *merge*, not an edit — refused with a 409 that points at the golden-record merge, and the report stays pending. Both the wine change and the resolution are audited (`admin.wine.update` with `via: 'wine-report'` + `from`/`to`); the reporter still earns Cellar Cred.

**The public wine page report link** — lazy-loaded modal, same pattern as the bottle page, shown to logged-in users next to the wishlist action.

GDPR: `WineReport` already deletes with the reporter's account (`userDataRegistry.js:508`); the new fields ride in the same document — nothing new to register.

## Testing

- New `wineReports.apply.test.js` (4 route-level tests over a live express server, matching the house harness): apply-with-hooks + reindex + dual audit, type-doesn't-touch-identity, the 409 merge-hint with the report left pending, and refuse-apply-without-suggestion / plain-resolve-never-loads-the-wine
- 513 backend tests across all route/model suites; 801 frontend; en-only strings

## Campaign complete

R1 producer canonicalization (#855) · R2 appellation system (#858) · R3 region queue (#859) · R4 importer invariants (#855) · R5 admin tooling UIs (#857) · R6 weekly health cron (#856) · R7 this PR — plus MCP curation parity, the force-reindex button, and the scan-bench matching fixes. Deploy-day runbook lives in the strategy doc: seed script → unify script (dry-run first) → incremental embed → reindex after any restore.

### DIFF
--- a/backend/src/mcp/tools/adminRegistry.js
+++ b/backend/src/mcp/tools/adminRegistry.js
@@ -161,11 +161,14 @@ registerTool({
   scope: 'read',
   requireRole: ['admin'],
   annotations: { readOnlyHint: true, openWorldHint: false },
-  inputSchema: {},
-  handler: async (_args, _ctx) => {
+  inputSchema: {
+    limit: z.number().int().min(1).max(100).default(30).describe('Newest first; the backlog can grow between reviews'),
+  },
+  handler: async (args, _ctx) => {
     const { listPendingRegions } = require('../../services/taxonomyReview');
-    const items = await listPendingRegions();
-    return ok(`${items.length} region(s) awaiting review`, items.map(r => ({
+    const all = await listPendingRegions();
+    const items = all.slice(0, args.limit || 30);
+    return ok(`${all.length} region(s) awaiting review (showing ${items.length})`, items.map(r => ({
       region_id: r._id,
       name: r.name,
       country: r.country,
--- a/backend/src/models/WineReport.js
+++ b/backend/src/models/WineReport.js
@@ -29,6 +29,20 @@ const wineReportSchema = new mongoose.Schema({
     type: mongoose.Schema.Types.ObjectId,
     ref: 'WineDefinition'
   },
+  // Optional STRUCTURED correction (strategy 2026-07-29 R7): "this field
+  // should say this" instead of prose the admin has to re-type elsewhere.
+  // Only the one-click-appliable string fields — region/country are refs and
+  // stay free-text in `details`. Resolution can apply this in one action.
+  suggestedField: {
+    type: String,
+    enum: ['name', 'producer', 'appellation', 'type'],
+    default: undefined
+  },
+  suggestedValue: {
+    type: String,
+    trim: true,
+    maxlength: 200
+  },
   status: {
     type: String,
     enum: ['pending', 'resolved', 'dismissed'],
--- a/backend/src/routes/admin/import.js
+++ b/backend/src/routes/admin/import.js
@@ -492,8 +492,25 @@ async function finalizeMintedWines(ids) {
       }
       // save(): the pre-validate hook recomputes canonicalKey the same way
       // (idempotent) and the slug hook skips non-new docs, which is why the
-      // slug is assigned by hand above.
-      await doc.save();
+      // slug is assigned by hand above. The slug probe is check-then-set, so
+      // a concurrent import can win the race — on a unique-index collision,
+      // bump the suffix and retry instead of abandoning the row unfinalized
+      // (audit 2026-07-29 F3: an abandoned row is exactly the invariant hole
+      // this function exists to close).
+      let attempt = 0;
+      for (;;) {
+        try {
+          await doc.save();
+          break;
+        } catch (err) {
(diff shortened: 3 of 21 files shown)


jagduvi1:
## Pre-release audit of the full campaign diff — remediated in the latest commit

Seven parallel slice auditors over `v1.92.1+somm → this branch` (42 files, +2.6k lines), one per PR wave plus a security/GDPR cross-cut, every finding hand-verified before fixing. **Result: 2 High, 6 Medium, 5 Low — all fixed here; 1 finding refuted on verification; the security cross-cut was clean on every axis (authz, injection incl. prototype-pollution, DoS, audit coverage).**

### The two Highs (both #858-related, both "old data meets new code")

1. **The producer-is-a-place gate wasn't using the appellation key fold.** My patch for it was *lost between edit and commit* — only the auditor's git evidence caught the discrepancy between the commit message and HEAD. Consequence: hyphenated appellation docs would never match the gate, silently reopening the place-as-producer hole (#842's fix class) for exactly the appellations the seed script promotes. The gate now queries **both** folds.
2. **Legacy Appellation docs are unmatchable.** Docs created before #858 carry old-fold keys; nothing backfilled them, so they'd sit in the taxonomy while their own wine strings kept appearing "unmatched" — and a seed re-run could mint a competing twin. New `scripts/backfill-appellation-keys.js` (dry-run default) re-derives every doc's keys with a same-country collision guard that skips + reports rather than silently overwriting. **Add to deploy-day: run it right after the seed script.**

### Mediums, fixed
- **Report-apply now normalizes** — the one-click apply ran none of the write-pipeline normalizers, so an applied suggestion could reintroduce tier-suffixed appellations / trailing vintages; it now runs the same field-specific normalizers, and also **re-indexes the bottles' denormalized search docs** (parity with the admin edit route).
- **Stale suggestion can't ride a reason switch** — cleared client-side on change, refused server-side (`suggestion ⇒ reason must be wrong_info`).
- **Force-reindex now waits for the Meili tasks** — `addDocuments` is 202+async; previously "done" meant *enqueued* and async failures were invisible. `syncViaCursor` now returns task uids and the route awaits them; boot-time callers unchanged.
- **Unify script tie-break** — a quarantined row's ancient `createdAt` could flip the majority winner against `producerSpelling.js`'s identical rule; `oldest` now only seeds from voting rows. Its key-drift check also computes both sides fresh (a stale stored key from an older algorithm caused false skips).
- **Importer slug race** — a concurrent-import `E11000` abandoned the row unfinalized (the exact invariant hole R4 closes); now retries with a bumped suffix.
- **GDPR export** — the report text a user types (`details`/`suggestedField`/`suggestedValue`) now flows into `GET /api/users/me/export`; deletion was already complete.

### Lows, fixed
Resolver consensus requires **all** matching docs to agree (limit-2 could miss a third dissenter) + deterministic sort + 200-char re-cap on adoption; appellation name length enforced on REST (MCP already had it); taxonomy usage counts exclude quarantined rows; REST region-approve made idempotent like its MCP twin; `list_region_review_queue` gains the standard limit; dead `instanceof Map` branch removed from the health job.

### Refuted on hand-verification
"Merge orphans `WineReport.wineDefinition`" — **false**: `reassignWineRefs` already re-points it (`wines.js:1563`); the auditor saw only the `duplicateOf` line. No change made.

### Deferred with reasons (not bugs, noted)
Concurrent-reindex lock (idempotent upserts, no corruption); nginx 120s proxy timeout on reindex at 10x scale (latent, tracked); dup-guard not scoped per suggested field (conservative behavior kept).

### Verification
**Entire backend suite: 2,421 tests / 161 suites green** (first full-suite run of the campaign) · 801 frontend green · backfill script dry-run validated against the prod-copy stack.

PR #884, opened 2026-08-03
fix(registry): stop identify-text minting a wine for every AI guess

jagduvi1:
## The bug

`POST /api/wines/identify-text` called `findOrCreateWine` **unconditionally**, writing a WineDefinition (plus Country/Region/Grape taxonomy) into the **shared registry** for whatever the model guessed — before the user confirmed anything and before any bottle existed. Both AddBottle and AddToWishlist also fired it **automatically alongside every ordinary search**, so simply searching polluted the registry and spent AI budget.

Measured on prod 2026-08-03: **85 of 162 `createdVia:'ai'` rows have zero bottles (52%)**, against 11/319 (3%) for `'ui'`. One user hunting for the Ballarin "Giuli" on 30 July left behind `[Colonnara] "Ballarin"`, `[Giuli] "Ballarin"`, `[Ballarin] "Ballarin"`, `[La Stoppa] "Giulia"` and `[Château La Verrerie] "Guili Guili"` — none with a bottle. This is the upstream source of most producer-split / wrong-producer tickets from the maturity-queue curation sessions.

## Backend

**identify-text is now read-only.** It probes with the existing `{ matchOnly: true }` and returns one shape with four states:

```
{ identified: {name, producer, country, region, appellation, type, grapes[], confidence} | null,
  match:      { wine } | null,
  candidates: [{ wine, score }],
  reason:     string | null }
```

There is deliberately **no top-level `wine`/`created`** — the rename makes any consumer that assumed a saved document fail loudly rather than dereference `undefined`.

- **Model output is normalized at the route** (field caps, wine-type allowlist, ≤20 grapes, confidence clamped). `findOrCreateWine` trims name/producer before consulting any option, so a `"name": 42` response used to 500 inside the probe.
- **The budget handling is split into phases.** The old single try/catch refunded on *any* throw — and a soft-zone probe crashed on `wine.toObject` when `wine` was null, so **every near-match AI identify both 500'd and reversed a completed billable call**, including its draw on the site-wide kill switch. Transport failure before the call still refunds; a probe failure does not.
- `find-or-create` accepts `source: 'ai'` so accepted suggestions stay measurable, and gains the missing **region** and **grapes** caps now that it is the only user-facing registry-write path.
- **`requireNonDemo` deliberately not added** to identify-text: `tryDebitAi` already denies demo accounts before Claude with a *coded* 403 (`demo_ai_disabled`), and the middleware would swap it for an uncoded one. A test pins this so nobody "hardens" it later.

## Frontend

- The automatic AI call is **gone from both search handlers**. AI is opt-in via the existing "Can't find your wine?" row — so library results now paint immediately instead of waiting behind a slower AI call.
- The AI card is **shape-aware** (saved wine = populated refs; suggestion = plain strings — reading `.name` off a string rendered silent blanks) and says whether the wine is already in the registry or will be added on confirm.
- **Accepting an unsaved suggestion routes through `find-or-create`**, which brings the soft-zone *"did you mean one of these?"* prompt to a path that never had it. This is the part that actually stops duplicates.
- Registry results are **no longer hidden behind the AI card** (that hiding is how a guess beat the correct row), with near-matches de-duplicated.
- The accept button has a **busy state** — it is now an async write, and a double-tap could otherwise mint two rows.
- **WinePicker** (journal pairings) loses its registry write entirely: a pairing needs no registry id, and it fired from free-text dinner notes with no confirmation step, often before the entry was saved. It now resolves to an existing wine or just keeps the name, and reports failures instead of silently doing nothing.

## For Weblate reviewers

i18n is **EN-only**. Three keys were **reworded in place** (keys kept, so sv/fr/de retain translations but now carry stale semantics):

| Key | Was |
|---|---|
| `addBottle.aiFoundWine` | "Match found" |
| `addBottle.searchHint` | "…we'll identify and add the wine." |
| `addToWishlist.aiFound` | "AI found this wine" |

## Tests

New `wines.identifyText.test.js` (12) and `wines.findOrCreate.test.js` (9 — that route had **no** coverage), 5 new `findOrCreateWine` `matchOnly` cases including the soft-zone shape identify-text hits most often, and `WinePicker.test.js` (5).

`wines.aiBudget.test.js`'s "successful identify" asserted a `{ created: true }` fixture that is **impossible** under `matchOnly` — a green that would have survived a completely unfixed route; it is now reduced to its budget claim.

Green: 555 backend tests across 38 wine/mcp/auth suites, 113 more across import/bottles/cellarImport, 823 frontend tests, plus a clean `npm run build`.

## After deploy — expected, not regressions

- identify-text 500s drop to ~0 (the soft-zone crash is fixed).
- Daily AI debits tick **up** slightly (those calls are no longer wrongly refunded) while total AI calls drop sharply (no more per-search identification).
- `SimilarWinesModal` shown-rate goes up — that is the anti-duplicate prompt doing its job.
- `createdVia:'ai'` **changes meaning**: before this release it meant "auto-minted, nobody confirmed" (~162 rows); after, "a human accepted an AI suggestion". The `createdAt` deploy boundary separates the cohorts.

## Follow-ups (not in scope)

- `import.js` discards soft-zone candidates.
- Converge `scan-label` onto `matchOnly`.
- Stop sending the discarded `labelImage` to find-or-create, *then* lower the 5 MB body carve-out in `app.js`.

**Compose impact: none.**

### DIFF
--- a/backend/src/routes/wines.aiBudget.test.js
+++ b/backend/src/routes/wines.aiBudget.test.js
@@ -243,18 +243,21 @@ describe('POST /api/wines/identify-text — daily AI budget', () => {
     expect(identifyWineFromQuery).not.toHaveBeenCalled();
   });
 
+  // Budget claim only — the response SHAPE is pinned in wines.identifyText.test.js.
+  // This used to assert res.body.wine.name against a { created: true } fixture,
+  // which is impossible under matchOnly and would have stayed green against a
+  // route that still minted.
   test('successful identify debits once', async () => {
     identifyWineFromQuery.mockResolvedValue({
       data: { name: 'Fixin', producer: 'Albert Bichot', country: 'France', grapes: [] },
       debugRaw: 'raw',
       debugReason: null,
     });
-    findOrCreateWine.mockResolvedValue({ wine: { _id: 'w1', name: 'Fixin', producer: 'Albert Bichot' }, created: true });
+    findOrCreateWine.mockResolvedValue({ wine: { _id: 'w1', name: 'Fixin', producer: 'Albert Bichot' } });
 
     const res = await postJson(app, '/api/wines/identify-text', { query: 'Albert Bichot Fixin 2019' });
 
     expect(res.status).toBe(200);
-    expect(res.body.wine.name).toBe('Fixin');
     expect(userDebits()).toBe(1);
   });
 
@@ -262,7 +265,7 @@ describe('POST /api/wines/identify-text — daily AI budget', () => {
     identifyWineFromQuery.mockResolvedValueOnce({ data: null, debugRaw: 'x', debugReason: 'exception: boom' });
     let res = await postJson(app, '/api/wines/identify-text', { query: 'whatever' });
     expect(res.status).toBe(200);
-    expect(res.body.wine).toBeNull();
+    expect(res.body.identified).toBeNull();
     expect(userDebits()).toBe(0); // refunded
 
     identifyWineFromQuery.mockResolvedValueOnce({ data: null, debugRaw: 'x', debugReason: 'ai_unknown: no idea' });
--- a/backend/src/routes/wines.findOrCreate.test.js
+++ b/backend/src/routes/wines.findOrCreate.test.js
@@ -1,0 +1,189 @@
+/**
+ * POST /api/wines/find-or-create — the registry-write chokepoint.
+ *
+ * Since identify-text became read-only, this is the ONLY user-facing path that
+ * mints a WineDefinition, and it now receives machine-generated payloads (the
+ * AI suggestion the user accepted, forwarded verbatim by AddBottle /
+ * AddToWishlist). It had no route-level coverage at all; these tests pin the
+ * guards and the provenance stamp.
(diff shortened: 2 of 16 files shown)


jagduvi1:
### CodeQL note

The initial run flagged `js/polynomial-redos` as "1 new alert in code changed by this pull request". It is **alert #199, open on `main` since 2026-07-18** — `backend/src/utils/normalize.js`, a file this PR did not touch. CodeQL attributed a pre-existing alert to the PR.

Rather than merge past a red check, it is fixed here (second commit): `resolveGrapeName` stripped a trailing blend percentage with `/\s*\d{1,3}\s*%$/`, which on a long run of spaces retried from every index and rescanned the run each time — quadratic on input that is **not** trusted (grape names arrive from label scan, import files and AI output). Bounded to `{0,3}`, which changes no legitimate input since a real percentage carries at most one space per side. Timing regression test added next to the existing percentage cases.

This closes one of the four deferred CodeQL Highs; the remaining three are the known false positives in `bottles.js`, `journal.js` and `imageSanitizer.js`.

PR #893, opened 2026-08-03
fix(security): audit 2026-08-03 remediation (revised after adversarial review)

jagduvi1:
Six of the eight findings from the `v1.94.0..v1.99.0` audit (5 parallel reviewers). Full report: `SECURITY_AUDIT_2026-08-03_v199.md`.

**Not in this PR:** H-1 (`/api/import/validate` still mints registry rows from unconfirmed AI guesses) — that's a real change moving creation into `/confirm`, and it deserves its own review. L-3 (brace-expansion 2.1.4) waits on settling whether the finding contradicts the 30 July "advisory metadata stale" dismissal.

## M-2 — prompt injection could permanently evade the review queue

The sharpest finding. A model steered to answer `confidence: 1.0` made its own row invisible to `GET /api/admin/wines/low-confidence`, whose only filter is `$lte: threshold`. Confidence was stored with a `typeof` check and **no clamp**.

The lever was real: `suggestProfile` substitutes `name/producer/region/appellation/grapes` into the prompt unescaped — and unlike `name`/`producer`, **`region` and grape names were stored without collapsing interior whitespace**, so a newline could pose as a new instruction, up to ~4 kB of it. Enrichment fires automatically on bottle add and writes back to the **shared** registry.

Three layers, because one wouldn't hold: confidence is clamped; substitution strips control characters, collapses whitespace and bounds length; and the mint chokepoint normalises region and grape names so the registry can't store multi-line taxonomy at all.

## M-1 — the sole user-facing registry-write path had no audit entry

`POST /api/wines/find-or-create` mints into the shared registry, and since identify-text went read-only it's where AI suggestions land once a user accepts them. It wrote no `AuditLog` while `admin/wines.js` and both MCP surfaces do. Same action name (`wine.create`) so the three read as one stream. `createdVia` doesn't substitute — it's client-asserted and later merges rewrite it.

## L-4, L-5 — caps in the wrong place

- The enrichment sanitisers capped array **length** but not element length, so unbounded model output reached the shared registry, the embedding text and the MCP tools — while a curator editing the same field was held to 40/60/1000. Bounds now come from `LIST_FIELDS` rather than being restated, so the two surfaces can't drift again.
- The `region`/grape caps lived at the route, and **both import callers bypass it**. Moved into `findOrCreateWine`, where every path goes.

## L-2 — a request that hung forever

`/find-or-create` was a bare `async` handler with `?.trim()` outside the `try`, so `{"name": 1}` threw a rejection Express 4 never catches: no response, socket held open, one log line. Wrapped in `asyncHandler`, checks made type-aware. The new `grapes` guard had the mirror-image gap — `typeof g === 'string' &&` let a non-string through to a `.trim()` 500.

## L-1 — mine, from #888

The blog filter rendered the raw `?tag=` value as a pill, so `?tag=<anything>` placed attacker-chosen text among the real tags. React escapes it — never XSS — but a convincing spot for "⚠ Verify your account". Now filtered to known tags, and the rule is extracted to `visibleTagsFor` so it's pinned by tests rather than by reading the page.

## Worth flagging

**Writing the confidence test caught a bug in the fix.** `Number(null)` is `0`, so my first version fabricated a confidence of `0` for a field the model never answered — a different lie from the one being prevented, and one that would have quietly filled the review queue with rows that had no confidence at all.

## Tests

Backend **168 suites / 2543 tests** (+10), frontend **49 / 855** (+8), production build clean. New coverage pins the clamp, the per-element truncation, the description bound, and the tag filtering — including that an 8 kB `?tag=` is dropped rather than rendered.

### DIFF
--- a/backend/src/routes/admin/import.js
+++ b/backend/src/routes/admin/import.js
@@ -2,7 +2,7 @@ const express = require('express');
 const multer = require('multer');
 const { parse } = require('csv-parse');
 const { requireAuth, requireRole } = require('../../middleware/auth');
-const { generateWineKey, generateWineSlug, normalizeString, normalizeAppellation, resolveCountryName, isRecognizedCountry, isUnknownName } = require('../../utils/normalize');
+const { generateWineKey, generateWineSlug, normalizeString, normalizeAppellation, resolveCountryName, isRecognizedCountry, isUnknownName, sanitizeTaxonomyName } = require('../../utils/normalize');
 const { canonicalizeWineName } = require('../../utils/producerPrefix');
 const { computeCanonicalKey } = require('../../utils/wineIdentity');
 const WineDefinition = require('../../models/WineDefinition');
@@ -192,7 +192,7 @@ async function getOrCreateRegion(name, countryId, userId, cache) {
     { country: countryId, normalizedName: normalized },
     {
       $setOnInsert: {
-        name: name.trim(),
+        name: sanitizeTaxonomyName(name),
         normalizedName: normalized,
         country: countryId,
         createdBy: userId,
--- a/backend/src/routes/admin/taxonomy.js
+++ b/backend/src/routes/admin/taxonomy.js
@@ -1,7 +1,7 @@
 const express = require('express');
 const mongoose = require('mongoose');
 const { requireAuth, requireRole } = require('../../middleware/auth');
-const { normalizeString, normalizeAppellation, normalizeAppellationKey } = require('../../utils/normalize');
+const { normalizeString, normalizeAppellation, normalizeAppellationKey, sanitizeTaxonomyName } = require('../../utils/normalize');
 const Country = require('../../models/Country');
 const Region = require('../../models/Region');
 const Grape = require('../../models/Grape');
@@ -268,7 +268,10 @@ router.post('/regions', async (req, res) => {
     hierarchy.push(name);
 
     const region = new Region({
-      name: name.trim(),
+      // Bounded like every other write path — the display name is stored
+      // separately from normalizedName and was only trimmed, so it could hold
+      // newlines and unbounded length (security audit 2026-08-03).
+      name: sanitizeTaxonomyName(name),
       normalizedName,
       country,
       parentRegion: parentRegion || null,
@@ -314,7 +317,9 @@ router.put('/regions/:id', async (req, res) => {
     }
 
     if (name) {
-      region.name = name.trim();
+      // Rename is the path that made a create-only guard pointless: a region
+      // minted correctly could be renamed to anything afterwards.
+      region.name = sanitizeTaxonomyName(name);
       region.normalizedName = normalizeString(name);
     }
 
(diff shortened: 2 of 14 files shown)


jagduvi1:
## Revised — an adversarial pass found three of these fixes didn't work

Two more reviewers went over the remediation itself, told to assume the author was careless. They were right to.

### Confidence clamp — was cosmetic

The queue filters `{ $ne: null, $lte: threshold }`, so **null is excluded**. Sanitising unusable model output to `null` routed every bad value into the one state guaranteed to hide the row. Before the fix a bogus confidence hid it; after, it still hid it. I closed nothing.

Reachable for real: `JSON.parse('{"confidence":1e400}')` yields `Infinity` without throwing.

Fixed at the **query** rather than the value, in both the admin queue and the registry watchdog — null now matches, because a row nobody can vouch for is the row most worth reviewing. Prod has **2** such rows already invisible, which no change to the write path would ever have reached.

Kept `null` rather than fabricating `0`: `0` is a real value and inventing it is its own lie. Verified only 2 rows surface, so no queue flood.

### Prompt sanitising — bypassed by two characters

`String.prototype.replace` honours `$&`, `` $` ``, `$'` and `$$` in the **replacement** even when the search pattern is a plain string. Confirmed:

```
'AAA{{name}}ZZZ'.replace('{{name}}', "$'")  →  "AAAZZZZZZ"
```

A wine named `$'` expanded to the entire remainder of the template — one well-formed prompt, then the attacker's trailing instruction, and a ~30× token bill from a single field. My sanitiser stripped control characters and never touched `$`.

Now `replaceAll` with a replacer **function**, which is inserted literally and honours no pattern. `replaceAll` because the template is superadmin-overridable and a custom one may use a placeholder twice.

### Taxonomy "chokepoint" — wasn't one

This was the conceptual error: I guarded **create** and called it a chokepoint while `PUT /regions/:id` could rename any row afterwards. Five bypasses in total. `sanitizeTaxonomyName` now lives in `utils/normalize.js` beside `normalizeString` and covers both admin create routes, **both renames**, and the admin import upsert.

### Also, from the regression pass

- The `{{grapes}}` cap applied to the **joined** string, truncating a large blend mid-varietal before it reached the prompt. Now per name.
- The tag filter compared **case-sensitively** while the server lowercases `req.query.tag` — so `?tag=Feature` filtered the list, dropped the pill *and* left nothing marked active. Precisely the failure the active-tag rule exists to prevent.

### Cleared, with reasoning

`normalizeString` **already** collapses whitespace, so the taxonomy change cannot split the registry by minting duplicates for legacy rows — that was the biggest risk in the first version and it doesn't exist. `MAX_FIELD` at module scope, the `asyncHandler` wrap, and the signature changes all checked clean.

### Tests

The confidence test **asserted the hiding as correct behaviour** — rewritten, along with the two query-shape assertions in `wines.lowConfidence.test.js`. New coverage pins the `$`-substitution rule and the case handling.

Backend **169 suites / 2551 tests**, frontend **49 / 860**, build clean.

PR #906, opened 2026-08-08
Add color inheritance to history section header

opr:
When accessing history, the heading isn't legible because it is using the default colour, and on a dark background does not contrast well. Inheriting the colour would fix this

Before: 
image

After:

image

### Testing

1. To test, add a bottle to your history, preferably one in each category (Drank, Gifted, Sold, and Other) and navigate between tabs.
2. Ensure the section title matches the section colour.

### DIFF
--- a/frontend/src/pages/CellarHistory.css
+++ b/frontend/src/pages/CellarHistory.css
@@ -184,6 +184,7 @@
 .history-section-header h2 {
   margin: 0;
   font-size: 1.1rem;
+  color: inherit;
 }
 
 .history-section-icon { font-size: 1.2rem; }


jagduvi1:
Thanks @opr — nice catch, and exactly the right fix. Verified the mechanism before merging: the global `h1, h2, h3 { color: var(--color-text) }` in `index.css` was beating the per-category header colours, so the Drank header's hardcoded dark background went dark-on-dark in light theme. `color: inherit` at the header's specificity is cleaner than duplicating the four category colours would have been.

Merged — it will be in the next release. The before/after screenshots and test steps made this a quick review, much appreciated. 🍷

PR #942, opened 2026-08-12
feat(registry): identity quality at the point of writing

jagduvi1:
Six items, in priority order. The first two came from defects found in live use hours after v1.107.0 shipped.

## P1 — Promotion must test that an identity is PLAUSIBLE, not merely non-empty
A wine with producer **"Increíble"** and name **"Increíble"** — the name echoed into the producer field — satisfied promotion, left the pending queue, and reached the maturity queue. Once promoted, its label photo was no longer reachable, so the one piece of evidence that could fix it was gone. The rule could only test "non-empty and not the word Unknown".

New shared predicate `isImplausibleIdentity`, synchronous and dependency-free (a pre-validate hook must never do DB I/O), wired at **three** call sites — the promotion hook, mint, and the curation write path — so one definition governs all of them. It catches the echo, one-character and digits-only producers, producers made only of house words ("Domaine"), and containment that leaves no distinct producer.

It deliberately does **not** judge non-Latin scripts — `normalizeString` can't represent them, so every fold would read "empty" and condemn a real producer. That's exactly the H-4 regression that made "Мукузани" rows unpromotable, and it stays fixed.

The taxonomy half (is this producer actually a place or a grape?) needs the DB, so it runs at the curation write path and **tells** the curator which rule fired rather than silently declining to promote. Only the six producer-is-not-a-producer rules are enforced there — `name-in-producer` would refuse a curator who transcribed "Cloudy Bay / Cloudy Bay Sauvignon Blanc" correctly off the label.

## P2 — A promoted wine's label scan stays readable for 7 days, then goes
Previously the scan was **retained indefinitely but unreachable** after promotion — the worst of both. Now promotion starts a 7-day clock; within it a curator can still correct a wrong completion against the label; after it the file, the document and the wine's pointer all go. Still-pending wines are on neither clock — their scan is the evidence. **This reduces retention**, and only the scan is served post-promotion; owners' bottle photos become private again. Side effect worth noting: the REST gate was *tightened* — it previously handed any label scan to any somm token.

## P3 — Two cross-field rules (sommelier Ticket 3)
`colour-contradiction.v1` — a colour term in the name contradicting `type` (live: Domaine Rolet "Rosé Poulsard" typed red). Only judged when `type` is itself a colour, so "Blanc de Blancs" sparkling never fires, and noir/nero are excluded since they name grapes. `cuvee-near-miss.v1` — a name highly similar but not identical to another wine from the same producer, the vision-misread signature (live: "Coeur de Roi" vs the real "Cœur de Roches"). Flags only, never auto-changes.

## P4 — Region adjectival synonyms
Script (dry-run default, **not run**) giving regions their adjectival label forms as synonyms, so the place-as-producer gate catches "Tokaji" when the region is "Tokaj". 28 entries, each justified; it refuses to alias a synonym another region already owns (that's a merge) and creates nothing.

## P5 — Slug correctness
A name correction didn't regenerate the slug: the Magnien wine, corrected this week to "Nuits-Saint-Georges Premier Cru Cœur de Roches", still lived at `…/coeur-de-roi`. Slugs now regenerate on rename with `previousSlugs` preserving every old URL (OG 301s to the current one). Sitemap and MCP were checked — neither resolves a wine by slug.

## P6 — "No producer on the label" disposition
Some bottles genuinely print no producer — two are in prod now. `identityUnavailable` takes the row out of the queue while the wine stays pending and unshared (promoting a producerless wine would wreck dedup — producer is 45% of the composite). Reversible, auto-cleared if the identity is later completed, and described in MCP as a **last resort after asking the owner** — which works on pending wines as of #941.

## Tests
Backend **222 suites / 3338 tests green** (from 216/3210). Frontend **56 files / 912 green**, run from the main checkout. Seven pre-existing tests updated where behaviour deliberately changed, each with a comment saying why.

## Honest limitation
`normalizeString` deletes the `œ` ligature rather than folding it to `oe`, so the near-miss rule scores the bare "Coeur de Roi"/"Cœur de Roches" pair at 0.54 and only trips on the longer stored names. Folding it would change `normalizedKey` for existing rows — a registry-wide migration, deliberately not attempted here.


jagduvi1:
## Audit remediation applied — the plausibility rule was rebuilt, not retuned

Two agents audited this; the backend one returned **DO-NOT-SHIP**, and it was right. `isImplausibleIdentity` blocked **67 of 91** sampled real-world producer/name pairs — Château Margaux, Petrus, Krug, Dom Pérignon, Opus One, Sassicaia, Tignanello, Château d'Yquem — because it treated *producer == name* as a defect. It isn't: the single-wine estate is one of wine's most prestigious categories. Stop-word stripping made it worse (Domaine de la **Romanée-Conti**/Romanée-Conti, **Harlan Estate**/Harlan), and the digits-only term condemned the real Penedès house **1+1=3**.

Shipped, that would have silently deleted the producer of the first Petrus anyone added and buried the wine in an invisible queue.

**The architecture changed, not the thresholds.** Identity quality cannot be judged from string shape:
- **Blocking** now covers only what cannot be an identity: a sub-2-character producer, or a producer that is entirely house/stop words and short (`Domaine`, `Casa`, `The Wine Estate`).
- *Producer echoes name* became a **flag** — new `producer-echoes-name.v1` in the cross-field queue. Château Margaux gets cleared in a click; "Increíble" gets fixed. The motivating bug is **surfaced rather than buried**, and P2's 7-day window keeps its label readable while a curator judges it.
- The famous-wine set is now a named regression suite so this class cannot return.

**Independently re-verified against the branch: 25/25 real pairs pass, 4/4 junk pairs still blocked, sentinels unchanged.**

Also fixed: the admin edit path promoted wines without starting the retention clock (leaving scans unreadable *and* unsweepable — exactly the state P2 removes); an import path could mint a slug already held as another wine's previous URL, so one URL matched two documents; `colour-contradiction` flagged **Cheval Blanc**; two cross-jurisdiction region aliases (Argentina's La Rioja, France's Moselle AOC) were removed; the cross-field write gate gained an audit-logged override so a bad taxonomy row can't make a real producer permanently unwritable.

**Honest disagreement worth reading** (`cuvee-near-miss`): the audit's premise was mis-measured — the Magnien pair scores 0.7333, not 0.80, and 0.5385 as stored because the normalizer deletes the `œ` ligature. No threshold reaches it without also flagging `Brunello di Montalcino`~`Rosso di Montalcino` (0.7273), two real wines many estates both make. The threshold moved to 0.78 with a numeric-token guard, and the measurements are recorded in the code rather than the gap being papered over.

Backend **3404 tests green**; frontend **912**.

PR #966, opened 2026-08-14
fix(registry): adopt the registry's producer spelling on every mint surface

jagduvi1:
## Why

The 2026-08-14 cleanup consolidated **130 producer display splits (584 rows)** — one estate stored as several strings (`Philipp Kuhn` / `Weingut Philipp Kuhn` / `Felton Road Wines Ltd`). This PR stops new ones from forming, **without changing the add or import experience in any way**: the fix is a silent spelling adoption that can only change which string is stored, never whether a save succeeds. On any doubt it keeps the user's input and leaves the rest to a periodic 10-minute cleanup session.

## Where splits came from

| Source | Hole | Fix |
|---|---|---|
| Scan / MCP / cellar import (via `findOrCreateWine`) | spelling adopter groups by `normalizeString`, which keeps estate titles → title variants never fold | second, country-fenced stage in `producerSpelling.js` |
| LWIN admin import (unused in prod — early test only) | drops `PRODUCER_TITLE`, bypasses the adopter entirely | resolves producer per row (cached); pre-resolve keys ride the upsert `$or` (F3 pattern) |
| Admin create + wine-request approval | build rows directly, no spelling resolution | both call the resolver |
| Weekly health watchdog | same `normalizeString` blindness — counted **zero** of the 130 clusters | `producerDisplaySplits` metric (producer key + country, sentinels excluded), new key → fresh baseline |

## The adoption rule (stage 2)

Adopt the registry's spelling **only when all three hold**, else keep the typed string:

1. **Same country** — `Weingut Jordan` (DE) never captures a new `Jordan Winery` (US) wine
2. **Exactly one existing spelling group** for that producer key — contested buckets (`Château de Seguin` vs `Chateau Seguin`, two real Bordeaux estates) adopt nothing
3. **Decoration-only difference** (token subset) — a typo (`Philip Kuhn`) or different wording (`Stags Leap Winery` vs `Stag's Leap Wine Cellars`) is never auto-adopted

Because a decoration adoption changes the producer's fold, the mint key is now computed from the **stored** spelling (the same-string stage was fold-invariant, so this is new only for stage 2).

## Also in this PR

- `scripts/consolidate-producer-displays.js` — yesterday's cleanup productized (dry-run default, `--apply` writes): fullest-estate-form target rule, held-cluster guards, `save()`-based writes so slugs regenerate with `previousSlugs` intact, unique-key collision pre-checks.
- `producerSegment` exported from `wineIdentity.js` so resolver + sweep + canonicalKey share one fold.

## Tests

261 tests across 14 suites green in node:20-alpine, including new coverage: stage-2 adopt (both directions), contested-bucket hold, typo hold, no-country hold, fail-open on DB error, mint-key-follows-stored-spelling, watchdog title-split counting / country fence / sentinel exclusion, upsert legacy-keys composition.

## Notes

- No compose/env impact. No MCP schema change (somm needs no fresh conversation).
- The watchdog's `producerDisplaySplits` appears as NEW METRIC once in the prod registry-watch output — expected, baselines on first run.
- Existing 22 held clusters (Taylors, Stags Leap, Seguin…) stay as they are by design; the sweep script reports them every run.


jagduvi1:
Added per review discussion: **`nearProducerPairs` now rides the weekly health job** (commit af04b36). Mint-time adoption deliberately refuses edit-distance folds (Philip/Philipp Kuhn — a 1-edit pair can be two real estates, Rockford/Rochford), so those land in the admin fragmentation queue — which nobody opens unprompted. Its total now alerts-on-increase like every other metric, sourced from the same function the admin page calls so the count and the queue agree. One more NEW METRIC line in the prod watchdog output on first run — expected.

PR #1147, opened 2026-08-30
feat(racks): add row-flip option for Honeycomb racks

khinsberger:
# Add row-flip option for Honeycomb racks

## Background

Honeycombe-style racks with an even number of rows can either be built up with the larger number of slots (N) on the top or bottom. The currently implemented default was that the row with N slots is always on top with no possibility to flip the rack layout top-to-bottom.

## Summary

- Adds `typeConfig.hexFlip` so a Honeycomb (hex) rack's alternating row-width sequence can be mirrored top-to-bottom, letting either the N-column or the N-1-column row sit on the bottom.
- This is a pure reversal of the existing row-width sequence — total slot count is provably unaffected. For an **odd** row count the unflipped sequence is always a palindrome (e.g. `N, N-1, N`), so the toggle is a visible no-op there by construction; for an **even** row count it correctly swaps which end is short.
- Threaded through the 2D SVG rack layout (`rackLayouts.js`), the 3D room view mesh (`RackMesh.js`), and a new "Flip rows (short row on top)" checkbox in the rack-creation form. No backend capacity/validation changes were needed since the total slot count never changes.

## Test plan

- [x] `npx vitest run src/utils/rackLayouts.test.js` — 64 passed, including two new tests covering the odd-row no-op and the even-row swap.
- [x] `npx jest src/utils/rackGeometry.test.js` (backend) — 45 passed, confirming totals are unaffected as expected.
- [x] `npm run build` (frontend, Vite) — compiles cleanly.
- [x] `npm test` for both frontend and backend — all tests passed.
- [x] Manually verified in a live `docker-compose` stack: created a flipped Honeycomb rack via the New Rack form, confirmed the preview/real 2D rack view and the 3D Room View all render the mirrored row sequence identically, with slot numbering and total count unchanged.

### DIFF
--- a/backend/src/models/Rack.js
+++ b/backend/src/models/Rack.js
@@ -26,7 +26,10 @@ const rackModuleSchema = new mongoose.Schema({
     moduleCols: { type: Number, min: 1, max: 10 },
     bottlesPerCell: { type: Number, min: 1, max: 20 },
     bottlesPerSection: { type: Number, min: 1, max: 30 },
-    backCols: { type: Number, min: 0, max: 20 }
+    backCols: { type: Number, min: 0, max: 20 },
+    // Hex racks only: mirror the row sequence top-to-bottom. Pure reversal —
+    // never changes total slot count (see rackSchema.typeConfig.hexFlip below).
+    hexFlip: { type: Boolean, default: false }
   },
   x:          { type: Number, default: 0 },
   y:          { type: Number, default: 0 },
@@ -58,7 +61,13 @@ const rackSchema = new mongoose.Schema({
     // Deliberately NOT on rackModuleSchema above — modular-rack modules
     // don't support double-height rows (out of scope; Mongoose strips the
     // key from module typeConfig).
-    doubleHeightRows: { type: [Number], default: undefined }
+    doubleHeightRows: { type: [Number], default: undefined },
+    // Hex racks only: mirror the row sequence top-to-bottom (row 0 becomes
+    // the old last row's width, etc). Pure reversal of the same multiset of
+    // row widths, so it never changes total slot count — for an odd row
+    // count the unflipped sequence is always a palindrome, so this is a
+    // no-op there by construction; it only matters for even row counts.
+    hexFlip: { type: Boolean, default: false }
   },
   // Modular rack fields (used when isModular is true)
   isModular: { type: Boolean, default: false },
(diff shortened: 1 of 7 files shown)


jagduvi1:
Thank you — this is a well-made contribution, and a real gap closed. A honeycomb rack built with the short row on top simply couldn't be represented before.

I checked the claims rather than taking them on trust, and they hold:

- **Total slots never change.** Ran the width sequences independently for 3–6 rows: `[5,4,5]`, `[5,4,5,4]`, `[5,4,5,4,5]`, `[5,4,5,4,5,4]` — reversal preserves the multiset every time.
- **Odd row counts are a genuine no-op.** The unflipped sequence is a palindrome, exactly as you say, so the toggle can't do anything there. Good instinct to pin that in a test rather than leave it as folklore.
- **Even row counts swap correctly** — `[5,4,5,4]` → `[4,5,4,5]`.
- **No backend change needed** — confirmed: `utils/rackGeometry` computes a slot *count*, never positions, so nothing server-side can disagree with the new layout.

The part that could easily have been missed and wasn't: every `computeLayout` caller already passes `rack.typeConfig`, so the flip flows through the rack renderer, print maps, the zone editor, auto-arrange geometry and the 3D room without extra plumbing — and you threaded `RackMesh` separately for the 3D mesh's own position function. That's the whole surface.

CI hadn't run (fork branches need a maintainer to approve workflows), so I ran the suites locally against your branch:

- `rackLayouts.test.js` — 64 passed, including your two new cases
- backend `rackGeometry` + `rackArrange` + `routes/racks` — 73 passed
- rack component tests — 14 passed
- CSS token guard — clean (`--color-text-muted` is a real token)

Merging.

---

One note for later, **not** a change request and not something I'd ask you to fix — it's a pre-existing gap your feature is simply the first to reveal:

`PUT /api/racks/:id` accepts `typeConfig` freely, and its only structural guard is the total-slot capacity check. `hexFlip` is the first `typeConfig` field that can *reorder positions* while leaving the total identical — so flipping an already-populated rack would pass validation and silently relocate every bottle relative to its slot number, making the user's cellar map wrong.

Your PR is safe as written precisely because the checkbox lives only on the **New Rack** form, and an empty rack has nothing to relocate. Restricting it there was the right call. If a flip toggle is ever added to the rack *edit* form, it'll need a guard (refuse when slots are occupied, or remap the stored positions). I'll open a separate issue for that.

Thanks again — clear description, honest test plan, and tests that pin the invariant instead of the implementation.

PR #1162, opened 2026-08-31
fix(mcp): handle pending wines in pairing tools

lukejduncan:
## Summary
- filter absent registry IDs before converting them to strings for WineDefinition queries
- share the ID normalization between `what_should_i_open_tonight` and `pair_with_dish`
- preserve pending-registry bottles in candidate and style-spread results with null taste profiles
- cover mixed, all-pending, idless, and equal-ObjectId deduplication cases

## Root cause
Pending-registry bottles legitimately have no `wineDefinition`. The previous code called `String()` before `filter(Boolean)`, turning an absent ID into the truthy string `"undefined"`. Mongoose then tried to cast that value inside an `_id: { $in: ... }` query and the entire tool call failed.

The backend was searched for sibling occurrences in the drinking-service/MCP path; both affected query builders now use the shared helper. A superficially similar analytics conversion is a separate data path and is intentionally outside this focused fix.

## Verification
- `npm test -- --runInBand src/mcp/drinkingTools.test.js` — 20/20 passed
- Node 24 CI-equivalent `npm test` — 282/282 suites, 4,532/4,532 tests passed
- `git diff --check`
- independent Opus code review: 10/10, no required changes

### DIFF
--- a/backend/src/mcp/drinkingTools.test.js
+++ b/backend/src/mcp/drinkingTools.test.js
@@ -78,6 +78,25 @@ const mkBottle = (over = {}) => ({
   ...over,
 });
 
+const mkPendingBottle = (over = {}) => mkBottle({
+  wineDefinition: null,
+  pendingWineRequest: { wineName: 'Pending registry wine' },
+  ...over,
+});
+
+function wineIdQueriesFromFindCalls() {
+  return WineDefinition.find.mock.calls
+    .map(([query]) => query?._id?.$in)
+    .filter(Array.isArray);
+}
+
+function expectWineIdQueries(expectedIds) {
+  const queries = wineIdQueriesFromFindCalls();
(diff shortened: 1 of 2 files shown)


jagduvi1:
Verified independently before merging: confirmed the bug is reachable (the candidate query has no wineDefinition filter, so every user with pending-registry bottles — exactly the people freshest from an import — could crash both pairing tools), confirmed the serializer and populate were already pending-aware so this is the whole fix, and ran your three pending-wine tests against the *unfixed* service to prove they fail there. Full MCP suite on your branch: 616/616 across 39 suites.

A model contribution: minimal diff, tests that pin the bug rather than the implementation, and an honest scope note about the analytics path you deliberately left alone. Merging as-is — this ships in v1.190.0 today. Thank you!

Issue #1203, opened 2026-09-02
Generalise the existing Google SSO to any OIDC provider


deltabeat:
### The problem

I self-host Cellarion and run my own identity provider (Pocket ID in my case), as a growing number of
self-hosters do — one sign-in covering every app on the network, with the
provider holding the credential rather than each app holding its own.

Cellarion is the only app I run that stays outside it. It offers a local
password or a Google account, and neither is my provider — so it's the one place
I need to keep a separate password. 

The machinery is there already, it's just pointed at Google specifically rather than at a
configurable issuer, so a self-hosted instance can't aim it at its own provider.

### What you would like

Generalise the existing Google SSO so the issuer is configurable, rather than
adding a separate mechanism.

Most of it already looks to be in place at v1.192.0:

- `authProviders` on the User model is a **generic array**; `provider` is
  `enum: ['google']` (`models/User.js:56`).
- Password is already optional for SSO-only accounts, gated on
  `authProviders.length` (`models/User.js:38`).
- Strategy registration is already **opt-in on config presence**
  (`GOOGLE_ENABLED = Boolean(GOOGLE_CLIENT_ID && GOOGLE_CLIENT_SECRET)`,
  `oauth.js:15`), and `/sso/providers` already tells the login page which
  buttons to render.
- `generateUniqueUsername()` and `upsertGoogleUser()` are provider-agnostic
  apart from the `'google'` literal and Google's `email_verified` shape.
- It's already passport, so `passport-openidconnect` would sit alongside
  `passport-google-oauth20` rather than replace it.

So roughly: add `'oidc'` to the enum, parameterise the upsert by provider,
register a second strategy behind `OIDC_ISSUER_URL` / `OIDC_CLIENT_ID` /
`OIDC_CLIENT_SECRET` following the same config-presence pattern rather than a
mode flag, plus the two routes and a conditional button. Google path untouched;
deployments without the new vars see no change.

**One design question that needs answering**
`upsertGoogleUser` refuses to link unless the email is verified
(`oauth.js:78`), and the comment explains why — an unverified address could
otherwise be used to take over an existing account sharing it. Clearly right for
Google. Generic providers vary: Pocket ID reports `email_verified: false` by
default unless an admin explicitly asserts verification, so a straight port would
refuse to link and surface as `no_verified_email`, which reads as "SSO is
broken". Either the provider must be documented as having to assert verified
[...]

Issue #1165, opened 2026-08-31
Signing in from a deep link lands on /cellars instead of the page you asked for


deltabeat:
If you open a link into the app while signed out, you're sent to `/login` — and
after signing in you land on `/cellars` rather than the page you were trying to
reach.

**Where I hit it**

I have NFC tags on the shelves of my wine fridge pointing at `/nfc/rack/<id>`.
On iOS a tag opens **Safari**, never the installed home-screen PWA — and iOS
gives a standalone web app its own cookie jar, so being signed in on the app
icon doesn't mean you are signed in in Safari. That Safari session is used *only* for
tag taps, which are far less frequent than the 7-day refresh cookie, so it has
nearly always expired by the next one.

The result, standing at the fridge: tap the tag → sign in → land on `/cellars`
→ walk back → tap the tag again. 

It isn't NFC-specific though — it's anyone following a shared or bookmarked link
into the app while signed out.

**Where it comes from**

`ProtectedRoute.js:25` redirects without recording where the user was headed:

```jsx
return <Navigate to="/login" replace />;
```

and `Login.js:80` always sends them to the same place:

```js
navigate('/cellars');
```

**Possible fix**

The usual react-router idiom — capture the location on the way out and honour it
on the way in. Roughly two one-line changes plus a `useLocation` import in each
file:

```jsx
// ProtectedRoute.js
const location = useLocation();
return <Navigate to="/login" state={{ from: location.pathname + location.search }} replace />;
```

```js
// Login.js
navigate(location.state?.from || '/cellars');
```

Storing the path as a string rather than the location object keeps the read on
the Login side to one optional chain and still preserves query strings.

One thing that might matter to you: because `state` is set by `ProtectedRoute`
itself and never read from a URL parameter, it can't be steered by a crafted
link the way a `?next=` query string could, so it needs no allow-list. Worth
saying since it touches the auth path.

I don't know whether the Google SSO round-trip should preserve the destination
too — that needs the target carried through the OAuth `state`, which is a bigger
change?

PR #1221, opened 2026-09-04
fix(auth): Google sign-in returns to the page you were heading for

### LINKED ISSUE #1220: Signing in with Google loses the deep link that #1165 restored for password login
### What happened

Following a link to a specific bottle or rack while signed out bounces you to
`/login`, as it should.

Sign in **with a password** and you land on the page you were heading for —
that's #1165, and it works.

Sign in **with Google** from that same screen and you land on `/cellars`. The
destination is gone.

The two paths diverge because they carry the destination differently, and only
one of them carries it at all. `ProtectedRoute` puts it in router state, which
survives a client-side navigation to `/login` but not a full-page redirect off
to Google and back. Nothing picks it up on the OAuth side:

| | |
|---|---|
| `frontend/src/components/ProtectedRoute.js:38` | sets `state={{ from: location.pathname + location.search }}` |
| `frontend/src/pages/Login.js:84` | password path reads it back: `navigate(location.state?.from \|\| '/cellars')` — **works** |
| `frontend/src/pages/Login.js:314` | Google button: `window.location.href = '/api/auth/google'` — full-page navigation, **router state dies here** |
| `backend/src/routes/oauth.js:139` | the redirect out to Google carries nothing of the sort |
| `backend/src/routes/oauth.js:29` | `successRedirect` is a constant, `${frontendBase}/login/callback` |
| `frontend/src/pages/OAuthCallback.js:28` | `navigate(user ? '/cellars' : '/login', { replace: true })` — hardcoded |

So by the time Google hands control back there is nowhere a target could have
been read from, and the callback has no choice but the constant.

The case your own comment in `ProtectedRoute.js` describes is exactly where
this bites — an NFC tag on a fridge shelf opens Safari rather than the
installed PWA, iOS gives the standalone app its own cookie jar, so that Safari
session has almost always expired by the next tap. A Google user walks back to
the fridge and taps again, same as before #1165.

Reproduced on a local `docker compose` build of `main` at v1.200.2, against a
throwaway Google OAuth client. Starting from `/wishlist` while signed out and
choosing *Continue with Google* lands on `/cellars`; the same journey using
email and password lands on `/wishlist`.

**The fix**, working locally: keep the destination on the client across the
redirect, in `sessionStorage` — scoped to the origin and the tab, so it survives
the trip out to Google and back. No backend change, and nothing in it names
Google, so a second provider inherits the behaviour.

Worth saying why not the two more obvious routes — I proposed the `state`
parameter myself in #1203, before reading that comment properly. Both a `?next=`
query parameter and the OAuth `state` parameter put the destination somewhere an
attacker can set it, which is exactly the property your comment in
`ProtectedRoute.js` is protecting when it says router state "can't be steered by
a crafted link, so it needs no allow-list". Holding it client-side keeps that
property: the only writer is the app's own code, and nothing is ever read from
a URL.

I'll send this as a PR. It's the same plumbing #1203 needs, so it lands first
and separately.

### Steps to reproduce

    1. Sign out (a fresh private window is easiest).
    2. Open a protected page directly, e.g. /wishlist — or a deep link such
       as /cellars/<id>/room?focusRack=<id>
    3. You are bounced to /login. Correct so far.
    4. Click "Continue with Google" and complete consent.
    5. You land on /cellars. The page from step 2 is lost.

    Contrast: run the same steps but sign in with email and password at
    step 4, and you land on the page from step 2 (#1165).

### Where are you running Cellarion

Self-hosted

### Version

1.200.2

### Browser and device

Not device-specific — reproduced in Chrome on macOS, but the destination is dropped in the callback route, so every browser is affected equally

### Logs or screenshots

```shell

```

deltabeat:
## What this changes

Fixes #1220. A signed-out deep link bounces to `/login` and, since #1165, signing
in with a password finishes the journey. Signing in with Google still lands on
`/cellars` — `ProtectedRoute`'s router state doesn't survive the full-page
redirect out to Google and back, and nothing downstream carries a destination,
so `OAuthCallback` has only its constant to offer.

The destination is now stashed in `sessionStorage` before leaving, and consumed
by `OAuthCallback` on success. Scoped to the origin and the tab, it survives the
cross-origin trip — the same mechanism the OIDC client libraries use to carry a
PKCE verifier across it.

**Deliberately not a `?next=` query parameter or the OAuth `state` parameter**,
which is worth explaining since I suggested `state` in #1203 before reading your
`ProtectedRoute` comment properly. Both would put the destination somewhere an
attacker can set, giving up exactly the property that comment relies on — "can't
be steered by a crafted link, so it needs no allow-list" — and turning the auth
path into an open redirect unless a validator is added. Holding it client-side
keeps the property: the only writer is the app's own code, and nothing is ever
read from a URL. The path validation in the new util is belt-and-braces rather
than the defence, and says so in a comment, so nobody later mistakes it for the
thing standing between us and an open redirect.

Two smaller decisions, both tested:

- **A failed or cancelled sign-in leaves the stash alone.** The usual next step
  is to try again, and clearing there would throw away the thing worth keeping.
  It is consumed only on success, so a stale destination can never steer a later
  sign-in.
- **No backend change, and nothing in it names Google.** A second provider — the
  OIDC work in #1203 — inherits the behaviour with no further plumbing. Happy to
  move it into `oauth.js` if you'd rather it lived server-side; you mentioned
  `oauth.test.js` as the file to extend, and this landed somewhere else.

Five files: a new `frontend/src/utils/postLoginRedirect.js` plus its tests, a new
`OAuthCallback.test.js`, six lines in `Login.js` and one in `OAuthCallback.js`.

## How I tested it

Both suites, and a local `docker compose up --build` at v1.200.2 against a
throwaway Google OAuth client — run **before** the change to watch the bug, then
after.

- **Before:** from `/wishlist` signed out → *Continue with Google* → landed on
  `/cellars`. The bug, observed.
- **After:** same journey → landed on `/wishlist`.
- **Plain sign-in unaffected:** Google sign-in with no deep link still lands on
  `/cellars`.
- **Password path untouched:** `/wishlist` → email + password → `/wishlist`.
- **Cancel then retry:** cancelled at Google's consent screen, took *Back to
  login*, signed in properly → still arrived at `/wishlist`.

Unit tests: 11 cases on the util (validator, round trip, consumed-once, poisoned
store, storage unavailable) and 5 on `OAuthCallback`. I checked they fail for the
right reason — reverting `OAuthCallback.js` to `main` fails exactly the two that
assert the new behaviour, and leaves the three asserting existing behaviour
passing.

No screenshots: nothing visual changed, only where you end up.

## Checklist

- [x] `cd frontend && npm test` passes — 1160 tests, 80 files
- [x] `cd backend && npm test` passes — 4845 tests, 308 suites (untouched, run to
      confirm no collateral)
- [x] Smoke-tested in Docker (`docker compose up --build`) and used the change in
      the app
- [x] Commits are signed off (`git commit -s`, see CONTRIBUTING.md)
- [x] No non-English `translation.json` files touched (no strings added at all)
- [ ] n/a — no personal data stored, processed or shared. The stashed value is a
      path, held in the visitor's own browser and cleared on use.
- [x] If AI tools helped write this, that is mentioned above and I have run the
      result myself

**AI assistance:** written with Claude Code. I ran the result myself — both
suites, and the full before/after journey in Docker described above, including
watching the bug happen on unmodified `main` first. The design decision to hold
the destination client-side rather than in `state` was taken deliberately - if you disagree, please let me know.

### DIFF
--- a/frontend/src/pages/Login.destination.test.js
+++ b/frontend/src/pages/Login.destination.test.js
@@ -1,0 +1,76 @@
+/**
+ * Where a successful password sign-in lands. Router state is the #1165 path and
+ * already worked; the stash covers coming back to /login after an SSO attempt
+ * that failed or was cancelled, where "Back to login" navigates without state.
+ */
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import Login from './Login';
+import { stashPostLoginRedirect, takePostLoginRedirect } from '../utils/postLoginRedirect';
+
+// t returns the key, or the inline English fallback where one is given.
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key, fallback) => (typeof fallback === 'string' ? fallback : key) }),
+  Trans: ({ children }) => children,
+}));
+
+const navigateMock = vi.fn();
+let routerLocation;
+vi.mock('react-router-dom', () => ({
+  useNavigate: () => navigateMock,
+  useLocation: () => routerLocation,
+}));
+
+const loginMock = vi.fn();
+vi.mock('../contexts/AuthContext', () => ({
+  useAuth: () => ({ login: loginMock, register: vi.fn() }),
+}));
+
+vi.mock('../hooks/useVersion', () => ({ default: () => null }));
+
+beforeEach(() => {
+  window.sessionStorage.clear();
+  navigateMock.mockReset();
+  routerLocation = { pathname: '/login', search: '', state: null };
+  loginMock.mockReset().mockResolvedValue({ success: true });
+  // The SSO-providers probe; the button itself is not under test here.
+  global.fetch = vi.fn().mockResolvedValue({ ok: true, json: async () => ({ google: false }) });
+});
+
+// The mode toggle carries the same label as the submit button, so target the
+// form's submit directly rather than by accessible name.
+const signIn = async () => {
+  fireEvent.change(screen.getByLabelText('auth.usernameOrEmail'), { target: { value: 'someone' } });
+  fireEvent.change(screen.getByLabelText('auth.passwordLabel'), { target: { value: 'hunter2' } });
+  fireEvent.click(document.querySelector('form button[type="submit"]'));
+};
+
+test('router state still wins — #1165 is untouched', async () => {
+  routerLocation.state = { from: '/cellars/abc/room?focusRack=def' };
+  render(<Login />);
+  await signIn();
+  await waitFor(() => expect(navigateMock).toHaveBeenCalledWith('/cellars/abc/room?focusRack=def'));
+});
+
+test('falls back to a stashed destination when the router state is gone', async () => {
+  // The shape after cancelling at the provider: "Back to login" navigates
+  // without state, so only the stash remembers where they were heading.
+  stashPostLoginRedirect('/wishlist');
+  render(<Login />);
+  await signIn();
+  await waitFor(() => expect(navigateMock).toHaveBeenCalledWith('/wishlist'));
+});
+
+test('a plain sign-in with neither still lands on /cellars', async () => {
+  render(<Login />);
+  await signIn();
+  await waitFor(() => expect(navigateMock).toHaveBeenCalledWith('/cellars'));
+});
+
+test('consumes the stash even when router state wins, so it cannot be inherited later', async () => {
+  routerLocation.state = { from: '/cellars' };
+  stashPostLoginRedirect('/wishlist');
+  render(<Login />);
+  await signIn();
+  await waitFor(() => expect(navigateMock).toHaveBeenCalled());
+  expect(takePostLoginRedirect()).toBeNull();
+});
--- a/frontend/src/pages/Login.js
+++ b/frontend/src/pages/Login.js
@@ -3,7 +3,7 @@ import { useNavigate, useLocation } from 'react-router-dom';
 import { useTranslation, Trans } from 'react-i18next';
 import { useAuth } from '../contexts/AuthContext';
 import useVersion from '../hooks/useVersion';
-import { stashPostLoginRedirect } from '../utils/postLoginRedirect';
+import { stashPostLoginRedirect, takePostLoginRedirect } from '../utils/postLoginRedirect';
 import './Login.css';
 
 const LOGO_WEBP = '/cellarion-logo-light.webp';
@@ -81,8 +81,14 @@ function Login() {
         setRegisteredEmail(result.email);
         setRegistered(true);
       } else {
-        // Finish the journey they started, or the default home for a plain sign-in.
-        navigate(location.state?.from || '/cellars');
+        // Finish the journey they started, or the default home for a plain
+        // sign-in. Router state is preferred; the stash covers coming back here
+        // after an SSO attempt that failed or was cancelled, where the state is
+        // gone. Consumed unconditionally — leaving it behind when router state
+        // wins is how an abandoned journey gets inherited by a later sign-in in
+        // this tab.
+        const stashed = takePostLoginRedirect();
+        navigate(location.state?.from || stashed || '/cellars');
       }
     } else {
       if (result.code === 'EMAIL_NOT_VERIFIED') {
--- a/frontend/src/pages/Login.js
+++ b/frontend/src/pages/Login.js
@@ -3,6 +3,7 @@ import { useNavigate, useLocation } from 'react-router-dom';
 import { useTranslation, Trans } from 'react-i18next';
 import { useAuth } from '../contexts/AuthContext';
 import useVersion from '../hooks/useVersion';
+import { stashPostLoginRedirect } from '../utils/postLoginRedirect';
 import './Login.css';
 
 const LOGO_WEBP = '/cellarion-logo-light.webp';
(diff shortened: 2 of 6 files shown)


jagduvi1 (changes requested):
Thanks — this is a well-argued fix and I'd like to merge it. The line
references check out, the test numbers match what I get on your branch, and
your reasoning for avoiding `?next=` and `state` is right. I've approved the
workflows so CI and DCO run.

Two things before merge:

1. **`OAuthCallback.js:32` — StrictMode double-run.** The stash is consumed
   inside a mount effect, and dev runs under `React.StrictMode`
   (`index.js:11`), which runs mount effects twice. Run one takes the
   destination and navigates; run two finds nothing and replaces with
   `/cellars`, which wins. So under `npm start` the fix never works
   (production is fine). Your test can't see it because it renders without
   StrictMode. We guard this exact trap in `VerifyEmail.js:22-25` with a
   `ranRef` — same shape here, plus a StrictMode-wrapped test asserting the
   last navigation. Depending on `Boolean(user)` instead of the object also
   stops a profile refresh re-firing the effect.

2. **The stash never expires.** `stashPostLoginRedirect(undefined)` skips
   rather than clears, and only an explicit logout wipes sessionStorage. So:
   cancel at Google → "Back to login" → password sign-in lands on `/cellars`
   with the destination still in storage; and a later plain Google sign-in
   in that tab is steered to it — on a shared computer, someone else's cellar
   URL. Clear the key when a sign-in starts with no destination, and let the
   password path fall back to the stash
   (`location.state?.from || takePostLoginRedirect() || '/cellars'`).

Optional, happy to do it in a follow-up instead: stash in `ProtectedRoute` at
bounce time and have both paths consume one store. Then every provider button
is a bare `window.location.href`, which #1203 will want. Not a condition.

Pre-existing and mine to fix: an ordinary user following an `/admin/*` link
lands on the bare "Access Denied" block after sign-in.

Thanks again — looking forward to the OIDC work.

deltabeat:
Both confirmed and both fixed in 36cbd0c — thank you, these were good catches.

**StrictMode.** You're right, and the uncomfortable part is that our smoke test
couldn't have found it: `docker compose up --build` is a production build, and
StrictMode only double-invokes in dev. So the verification I was most confident
in was structurally blind to the failure. `ranRef` guard as you suggested, same
shape as `VerifyEmail`, and the effect now depends on `Boolean(user)`. The new
test renders inside `StrictMode` and asserts the **last** navigation — asserting
a single call passes either way, which is exactly how this got through.

**Expiry.** Also right, and it falsifies a claim in my own commit message: I
wrote that a stale destination "can never steer a later sign-in". True for one
sign-in, false for the second. `stashPostLoginRedirect` now clears on an absent
or unsafe path, and the password route consumes the stash too.

One deliberate difference from your suggestion there. You proposed
`location.state?.from || takePostLoginRedirect() || '/cellars'`, which
short-circuits and leaves the stash in place when router state wins — so:
bounce to `/wishlist`, start Google, cancel, navigate to `/profile`, bounce,
sign in with a password. You land on `/profile` correctly, and `/wishlist` is
still in storage for whoever signs in next. I consume unconditionally and prefer
state, which is one extra line. Happy to take your version if you'd rather.

**On the validation, which I got wrong in the other direction.** The util's
comment claimed nothing is ever read from a URL and the path checks were
belt-and-braces. Both too strong — the destination derives from the path the
visitor was on, and an attacker picks that by picking the link. I convinced
myself it isn't exploitable: `//evil.example` matches no protected route and
falls to the catch-all at `App.js:542`, which isn't wrapped in `ProtectedRoute`
and passes no `from`. But that's the route table's doing rather than this file's,
and `@remix-run/router` falls back to `window.location.assign()` when `pushState`
throws on a cross-origin URL — so the checks are load-bearing the day a wildcard
route becomes protected. The comment now says that instead. Flagging it here
rather than privately because I don't believe it's reachable; tell me if you'd
rather it went through SECURITY.md and I'll move it.

**Verification this round:** both suites (frontend 1167, backend 4874), and a
dev run this time — `npm start`, destination seeded in `sessionStorage`,
`/login/callback` lands on it under StrictMode. Each of the three new failures
was reproduced before being fixed.

**The `ProtectedRoute` refactor** — I'd like to send it as its own PR rather than
grow this one, if that suits. It's the better end state and #1203 wants it, and
it's a different concern from the two bugs above. Worth noting one thing for
whenever it happens: stashing at bounce time means a destination is written on
every bounce, including journeys that never reach a sign-in, and it survives a
reload of `/login` where router state wouldn't. Same input, wider window — so the
clearing rules matter more there, not less.

And noted on `/admin/*` landing on the bare Access Denied block — leaving that
one to you.

jagduvi1 (approved):
Thanks for the revision — both points are addressed exactly as asked, and the commit message's own correction about the validation being load-bearing is right.

Verified on the branch: the callback runs its single-use consume once per mount cycle (`ranRef`, `Boolean(user)`), with a StrictMode-wrapped test asserting the last navigation; the stash now clears when a sign-in starts without a destination, and the password path consumes it unconditionally, so an abandoned journey cannot be inherited. Full frontend suite green here (81 files / 1167 tests).

Merging with today's release. The Access-Denied-after-SSO follow-up for role-gated links is still mine.

PR #1268, opened 2026-09-08
feat(auth): sign in with any OIDC provider, not only Google

### LINKED ISSUE #1203: Generalise the existing Google SSO to any OIDC provider

deltabeat:
Closes #1203.

A second strategy beside `passport-google-oauth20`, registered on config presence exactly the way `GOOGLE_ENABLED` is. No mode flag, no either/or — both providers can be on at once, and a deployment that sets none of the new vars sees nothing change. `OIDC_TRUST_EMAIL_VERIFIED` defaults off, as agreed on the issue.

## Choices worth your eye

**`passport-oauth2`, not a new dependency.** `passport-google-oauth20` is itself a `passport-oauth2` wrapper, and the cookie state store from GHSA-972f-r74x-7mfx implements that library's store interface — so the OIDC strategy inherits the browser binding and PKCE by passing the same `store`. A second provider therefore cannot reintroduce the gap that fix closed, which is why I waited for it to land.

**Claims come from userinfo; the ID token is deliberately ignored.** The token response carries one, because `openid` is in the scopes and providers refuse userinfo without it. Nothing reads it. An ID token is worth something only once its signature, issuer, audience and expiry are verified against JWKS, and the half-done version — decoding the JWT and believing it — is an authentication bypass. Using userinfo means there is no token to verify and no way to get that wrong. `nonce` is absent for the same reason.

**Six required variables, not one issuer plus discovery.** Strategies register at require time, so a discovery fetch would make sign-in depend on the provider being reachable at boot, and its absence would be silent. `OIDC_ISSUER` is required and used verbatim — deriving it from the authorization URL is unsound, because Keycloak appends `/protocol/openid-connect/auth` to its issuer, Pocket ID `/authorize`, Authentik `/application/o/authorize/`. Taking the origin would make every Keycloak realm on a host look like one issuer.

**Token-endpoint authentication is `client_secret_post`.** The credentials go in the token request body, not HTTP Basic — inherited from node-oauth, which writes them into the body unconditionally. Providers supporting both (Pocket ID, Keycloak, Authentik, Zitadel) need nothing; one registered for `client_secret_basic`, the OIDC default when discovery is silent, will reject the exchange. Authelia defaults to basic and needs `token_endpoint_auth_method: client_secret_post` on its client. Documented in `.env.example` and asserted by a test that inspects the real request. I did not implement basic because adding an `Authorization` header on top of body credentials sends two authentication methods on one request, so a real fix means replacing `getOAuthAccessToken` and duplicating its request building and response-parsing fallback here.

**It touches `User.js`.** `authProviders` entries gain an `issuer`, because OIDC guarantees `sub` unique only within an issuer. No migration: `oidc` has never shipped, so nothing in the wild holds an issuer-less link. Flagging it because it is slightly more than "a second strategy".

## Two pre-existing things this series fixes

**The linked-account lookup used two dotted conditions.** MongoDB matches those independently against the array, so an account holding `google/123` and `oidc/456` was returned for a query for `oidc/123` — an identity nobody linked — before the email checks. Harmless while one provider meant one entry; adding a second provider is what makes it reachable, so it is fixed here with `$elemMatch`. Reproduced against a live mongod, not reasoned about.

**The unit mock was hiding it**, implementing same-entry matching with `.some()`, which is stronger than the real database. It now reproduces both semantics faithfully.

## Testing

Backend 5237 / 344 suites, frontend 1260 / 98 (rebased onto v1.211.0). New coverage: the authorize redirect (state, PKCE `S256`, `openid` scope, state mirrored in the cookie), a callback refused before any code exchange without that cookie, a full callback through userinfo → claim mapping → account creation → session issuance, the token exchange itself (asserted below `getOAuthAccessToken`, so the request shape and the `code_verifier` are seen rather than stubbed over), identity-collision cases across providers and across issuers including two realms differing only by path, and configuration tests pinning that OIDC stays off without an issuer. Each was checked to fail for the right reason.

One limit worth stating: `issueTokens` is mocked in the route test, so it proves session issuance is reached with the right user, not that a usable session comes out. `authTokens` has its own suite, and the live run below is the end-to-end evidence.

## AI assistance, and having run it

Written with Claude, reviewed and run by me. It is not only read: I brought the stack up with `docker compose up --build` and signed in repeatedly against a real Pocket ID instance, which is how three defects were found that the suite could not have caught —

- the access token went to userinfo as a `?access_token=` query parameter, because node-oauth defaults `_useAuthorizationHeaderForGET` to false; the provider rejected it and the sign-in died as `server_error`
- the failure messages still said "Your Google account" to someone signing in with Pocket ID
- my own error text hid the first defect by reporting no HTTP status

Verified in the database afterwards: the account is keyed on the 36-character `sub`, not the email; `roles: ['user']`, so signing in through the IdP's admin group grants nothing inside Cellarion; the audit log records `auth.oauth.success { provider: 'oidc' }`; and repeat sign-ins are idempotent rather than accumulating provider entries.


jagduvi1:
Merged — thank you. This is a genuinely careful piece of work: reusing the state store so the second provider inherits the browser binding, refusing to half-verify an ID token, scoping the subject to the issuer, and catching the `$elemMatch` bug that only a second provider could reach. Running it against a real Pocket ID and finding the three things the suite could not was the part that made it easy to merge.

As promised on #1203, I'm pushing the small things on top rather than sending you round again: the button's translation key in the English catalogue (Google's was missing too), the OIDC block in `backend/.env.example`, and the README feature bullet. It ships in the next release.

One note for anyone reading later: GitHub's default CodeQL setup does not analyse pull requests from forks, which is why that check is absent here; it scans the merge commit on `main` instead.

PR #1302, opened 2026-09-12
feat(tokens): a personal API token can revoke itself (DELETE /api/tokens/self)

jagduvi1:
## Summary

Requested by the Home Assistant integration team: the integration cannot revoke its own API token when it is removed, because token deletion needs a session login plus the token id, neither of which it has.

- **`DELETE /api/tokens/self`** (`routes/tokens.js`, declared before `/:id`) — revokes exactly the calling personal token (`req.apiToken.id`, owner-scoped, `revokedAt: null`), audits `token.revoked` with `self: true`, and fires `eventBus.dropToken` (SSE streams + MCP sessions of that token die as with any revocation). A JWT session → 400 pointing at `DELETE /api/tokens/:id`; an already-revoked token → 404.
- **Scope allow-list** (`middleware/apiTokenAuth.js`) — `SELF_REVOKE` is checked before the per-scope rules, so any scope (read / consume / write / climate) may reach exactly this route. `/api/tokens` list, `/:id`, and `POST` stay session-only. OAuth-issued tokens remain confined to the MCP endpoint (unchanged).
- `docs/ha-push-events.md` gains the route.

## Tests

- `routes/tokens.test.js` (+3): self-revoke happy path (audit `self: true`, owner-scoped lookup), session → 400 without touching the model, revoked → 404
- `middleware/apiTokenAuth.test.js` (+2): every scope reaches `DELETE /api/tokens/self` (trailing slash too); list / other ids / `selfish` / `POST` / `GET self` stay denied

### DIFF
--- a/backend/src/middleware/apiTokenAuth.js
+++ b/backend/src/middleware/apiTokenAuth.js
@@ -111,11 +111,21 @@ function isMcpEndpoint(method, path) {
   return (method === 'POST' || method === 'GET' || method === 'DELETE') && /^\/api\/mcp$/.test(p);
 }
 
+/**
+ * The one route EVERY personal token may reach regardless of scope: revoking
+ * itself (DELETE /api/tokens/self — routes/tokens.js). An integration being
+ * removed holds only its own bearer token; letting it end that token is
+ * strictly narrowing, never widening. Exact-anchored; /api/tokens/:id and
+ * the token list stay session-only.
+ */
+const SELF_REVOKE = { method: 'DELETE', pattern: /^\/api\/tokens\/self$/ };
+
 /** Default-deny scope check. Exported for direct unit testing. */
 function isRequestAllowed(scopes, method, path) {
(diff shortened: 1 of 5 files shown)


jagduvi1:
Docker smoke on main after merge (`smoke-selfrevoke.mjs`), 8/8:

```
PASS POST /api/tokens mints a cel_ token — status=201 
PASS the token works (GET /api/auth/whoami) — status=200
PASS the token still cannot list tokens (scope denied) — status=403
PASS a session calling /self gets 400 pointing at /:id — status=400
PASS DELETE /api/tokens/self with the bearer → 200 — status=200 {"message":"Token revoked","id":"6aa568caee02f5cae0d4fb36","name":"Home Assistant (smoke 1789225161815)"}
PASS the token is dead afterwards (401) — status=401
PASS a second self-revoke is 401 (invalid token), not 200 — status=401
PASS the session token list no longer shows it as active — status=200

8/8 passed
```

